PR #2, opened 2015-09-09
Gradle RAT: Add plainOutput, xmlOutput and htmlOutput options

eskatos:
As requested in RAT-163, this pull-request contains changes to add RAT plain text output support.
BTW, xml & html output are now optional.

### DIFF
--- a/apache-rat-gradle/src/site/apt/index.apt
+++ b/apache-rat-gradle/src/site/apt/index.apt
@@ -66,6 +66,17 @@ rat {
     // Fail the build on rat errors, defaults to true
     failOnError = false
 
+    // Enable XML RAT output, defaults to true
+    xmlOutput = true
+
+    // Enable HTML RAT output, defaults to true
+    htmlOutput = true
+
+    // Enable plain text RAT output, defaults to false
+    // Please note that if xml or html output is enabled too,
+    // then two RAT runs will be needed to produce all reports.
+    plainOutput = false
+
 }
 -------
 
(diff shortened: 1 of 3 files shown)


apache-rat-gradle/src/site/apt/index.apt
```diff
@@ -66,6 +66,17 @@ rat {
...
 
+    // Enable XML RAT output, defaults to true
+    xmlOutput = true
+
+    // Enable HTML RAT output, defaults to true
+    htmlOutput = true
+
+    // Enable plain text RAT output, defaults to false
+    // Please note that if xml or html output is enabled too,
+    // then two RAT runs will be needed to produce all reports.
```
aw-altiscale: That's an extremely expensive requirement for large code bases.  We've seen with the RAT/gradle integration that the Samza team uses that it's definitely possible to generate multiple outputs from one run.

eskatos: The plugin uses the `report` ant task that runs rat and produces a single output.

`plainOutput = true` triggers a rat run and output plain text report
`xmlOutput = true` triggers a rat run and output xml report
`htmlOutput = true` triggers `xmlOutput` and xsl the xml output to html report

Defaults are respectively `false`, `true`, `true`.
If you only need the plain text output set all theses to their opposite.

However, it could be possible to get all reports with a single rat run by rewriting the plugin to use the RAT api directly. Or should we propose an enhancement for rat ant tasks?

eskatos:
This is now 3 years old, closing as outdated

PR #8, opened 2018-01-04
Stream backed writers passed to XMLWriter must use UTF-8 encoding

matthiasblaesing:
The XMLWriter writes a character stream and outputs 

<?xml version='1.0'?>

as XML declaration. The encoding declaration is missing, so encoding
must comply with UTF-8.

This change modifies the places where the output is written into a
file to use UTF-8 encoding. Without this change the default charset
of the platform is used, resulting in illegal bytesequences in the
output file.

### DIFF
--- a/apache-rat-plugin/src/main/java/org/apache/rat/mp/RatCheckMojo.java
+++ b/apache-rat-plugin/src/main/java/org/apache/rat/mp/RatCheckMojo.java
@@ -34,9 +34,12 @@ import org.apache.rat.report.claim.ClaimStatistic;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
-import java.io.FileWriter;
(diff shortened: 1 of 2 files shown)


ottlinger:
@matthiasblaesing thanks for your patch that seems to be related to 
https://issues.apache.org/jira/browse/RAT-242

As RAT is unfortunately not yet able to properly synchronize with Github (the repository is outdated and I already filed an INFRA-issue at ASF for that) I'll try to merge in the changes.

Do you already have an ASF-Committer's Agreement so that I can easily merge your changes in?!

Thanks for your help and interest in RAT!

ottlinger:
Applied to TRUNK via 
http://svn.apache.org/viewvc?rev=1820237&view=rev

Please verify that it fixes your problems and comment here or at RAT-242.

Thanks

PR #12, opened 2019-11-04
RAT-260: Use Docker in Travis CI to avoid animal-sniffer errors when building with various JDKs

mik-laj:
Hello

Travis has problems with depencies incompatibilities. The simplest solution is to use Docker, which provides isolation of dependencies. 
https://issues.apache.org/jira/projects/RAT/issues/RAT-260

Best regards,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,25 @@
-language: java
-jdk:
-  - openjdk8
-  - openjdk9
-  - openjdk10
-  - openjdk11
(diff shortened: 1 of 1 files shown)


ottlinger:
As you are in the ASF group already it's sure you signed a CLA - thanks for helping out :)

ottlinger:
Build is back in the green - thanks :+1:

PR #14, opened 2019-11-05
Add license header to .travis.yml

mik-laj:
Hello,

I noticed that the license header is missing in this file. I wonder if it's necessary, so I created PR to talk about it. What are your thoughts on this topic? Which files should have a header?

Best regards,

CC: @potiuk

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,21 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


potiuk:
Yeah. Shouldn't RAT repo be checked with RAT ;) ? Eating your own dog food :D

jochenw:
> Yeah. Shouldn't RAT repo be checked with RAT ;) ? Eating your own dog food :D

You are right, of course, Unfortunately, it's a non-trivial task, to build a Maven plugin, and use it within its own build. If you have any solutions to work around that chicken, and egg, problem: We re ready to listen.

potiuk:
Why not using CLI interface to check it ? Surely the .jar file is built at some point in time and you can just use 'java -jar SOURCES'

potiuk:
This way you can additionally check if the CLI interface still works.

potiuk:
If you think it might work, I might even try to dust off my maven shelf in the brain (been there done that) and contribute it ...

mik-laj:
I working on a PR now: https://github.com/mik-laj/creadur-rat/pull/2

mik-laj:
Final PR: https://github.com/apache/creadur-rat/pull/15

ottlinger:
Sorry for the misunderstanding - RAT checks itself during build:
https://github.com/apache/creadur-rat/blob/master/pom.xml#L246

Certain files are excluded as they are not part of the released artefact:
https://github.com/apache/creadur-rat/blob/master/pom.xml#L326

So we do not need a header in the Travis file.

PR #15, opened 2019-11-05
[RAT-261] Check the source code via Apache RAT to detect license violations

mik-laj:
Hello,

Release Audit Tool (Rat) is a tool to improve accuracy and efficiency when checking releases.  Rat was developed in response to a need felt in the Apache Incubator to be able to review releases for the most common faults less labour intensively. It is therefore highly tuned to the Apache style of releases.

Best regards,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
 *.ipr
 *.iws
 /target/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,39 @@ language: minimal
 services:
   - docker
 
+stages:
+  - lint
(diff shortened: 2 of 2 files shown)


ottlinger:
I'm sorry the whole discussion seems to be a misunderstanding.

RAT is checking itself during each maven build:
https://github.com/apache/creadur-rat/blob/master/pom.xml#L246

potiuk:
I think the whole discussion started with finding that the licence headers are missing in some of the files in RAT: .travis.yml, .gitignore, BUILD.txt for example. On the other hand it is in pom.xml, and ant-task-examples.xml. So we thought RAT does not use itself to check licences.

We had very similar problem in Airflow 1.10.6 rc1 release when one of the files which was not technically part of the sources of the app (was not released in package but it was released as part of the source.zip file) missed a licence. We actually dropped the release because of that and released 1.10.6rc2 with licence added. As a new PMC in Airflow I learned that it's my responsibility to not release software without licence headers.

So I think it is important to understand if this is intentional to skip those files from the check?

I looked through RAT documentation and the only thing I found is 

> useDefaultExcludes:
> 
> Whether to use the default excludes when scanning for files. The default excludes are:
> meta data files for source code management / revision control systems, see SourceCodeManagementSystems
> ...

It's true by default - but unfortunately there is no explanation what "SourceCodeManagementSystems" actually is (but I guess it is .git, .svn folders).

But there is no indication I could find that things like .gitignore or .travis.yml or BUILD.xml will also be ignored.

It would be great to understand the reasoning for those exclusions and possibly fix at least the .travis.yml and .gitignore and let RAT check that as well (RAT seems like perfect example that others will probably follow so better to have licence in .travis.yml)

ottlinger:
@potiuk sorry for the confusion - as I tried to tell you at ApacheCon. RAT is comprised of multiple modules:
* Maven support (mostly used and with the biggest range of functionality)
* Ant Support (added for downwards compatibility and for "older" projects
* CLI module

The CLI module does not allow to configure as much as the Maven integration.

As RAT itself uses the Maven integration it configures exclusions for scanning.

ottlinger:
.gitignore and .travis.yml are not part of the JAR / source-JAR and must not have a license header, therefore many projects configure it as excludes.

PR #18, opened 2020-05-13
[RAT-268] apache-rat-plugin incorrectly assumes poms module is always…

rfscholte:
… a directory


rfscholte:
Don't forget to close RAT-268 in Jira

PR #20, opened 2021-04-22
RAT-207: Properly report thread safety to Maven

kalgon:
AFAIK, the rat mojos seem to be thread-safe.

Nevertheless, If that change requires proper testing, I could write some tests (but I would need some guiding on how to do that).


ottlinger:
@kalgon thanks for reaching out to RAT.

@khmarbaise do you have any suggestions concering this PR? Is there anything else that needs to be done to test this change?

Thanks everybody :)

khmarbaise:
No you can't do more. The point is by using threadSafe only a hint that the plugin is being able to run during `mvn -T XXX` and the plugin does not use/need resources from outside which could be a problem in such cases.

khmarbaise (approved):

ottlinger:
@kalgon thanks for your contribution.

@khmarbaise thanks :)

PR #44, opened 2022-02-09
upgrade animal-sniffer-maven-plugin to 1.21

XenoAmess:


ottlinger:
@XenoAmess thanks for your PR - can you provide a bit more of explanation what this new profile does and how it may change the release process of RAT? By only seeing the changes I'd like to know more about any implications of the change.

XenoAmess:
well it should be able to pass ci but we don't have enough money for docker.

XenoAmess:
> @XenoAmess thanks for your PR - can you provide a bit more of explanation what this new profile does and how it may change the release process of RAT? By only seeing the changes I'd like to know more about any implications of the change.

@ottlinger
Fine.

1. javac added a parameter as **--release**, at jdk9+.
In short it describe which version of jdk would the produced jar used with, and openjdk guys will make sure it can run under it.
If not then fire bug report to them.
https://docs.oracle.com/javase/9/tools/javac.htm#JSWOR627
https://www.baeldung.com/java-compiler-release-option

2. maven-compiler-plugin can read maven.compiler.release property. If there be such a property, it will pass it to the javac.

3. **--release** is useful for jdk9+, but notice that if we pass this param to jdk8-, then it will fail error.
So we must find a way to only pass it when jdk version >= 9
That is why we use a auto-activate profile with condition.
Once the condition **<jdk>[9,)</jdk>** meat, means jdk version >= 9, then maven auto activate this profile.
When activate this profile, this profile set a property **<maven.compiler.release>8</maven.compiler.release>**
Then maven-compiler-plugin read the property **<maven.compiler.release>8</maven.compiler.release>** and pass it to javac.
Then the magics openjdk guys leaves in javac would help the compiler correctly compile jdk8-usable jar, which means do not choose jdk-8 unexist functions and other trickey things, that should not cared by we non-jdk repo.

ottlinger:
@XenoAmess you ran into the same problem as I did when I just updated the reference in the pom.xml alone:
https://app.travis-ci.com/github/apache/creadur-rat/jobs/558978072

ottlinger:
@XenoAmess can you be so kind to rebase your PR as the build configuration changed due to [RAT-296](https://issues.apache.org/jira/browse/RAT-296) ;) Did you already file a contributor agreement? https://www.apache.org/licenses/contributor-agreements.html 

Thanks again for your help

ottlinger:
@XenoAmess let's see if all pipelines run through .... I'll merge afterwards.

Thanks again for your help :) I'll add an entry to the changelog and do the red-tape later.

PR #135, opened 2023-04-08
RAT-272: run ci tests also with Java 17 and 20

slachiewicz:


ottlinger:
Thanks for raising the issue - due to the fact that RAT is Java8 compatible and these ancient versions do not play well with modern JDKs we only ran with the current setup.
Feel free to file an issue in ASF's Jira and start a discussion on the development mailinglist about changing the compiler level.

slachiewicz:
Current error with Java 17|

```
Error:  Generating /home/runner/work/creadur-rat/creadur-rat/target/site/apidocs/org/apache/rat/report/xml/writer/impl/base/XmlWriter.html...
Error: [ERROR] error: An internal exception has occurred. 
Error:    	(java.lang.NullPointerException: Cannot invoke "com.sun.source.util.DocTreePath.getTreePath()" because "path" is null)
Error:  Please file a bug against the javadoc tool via the Java bug reporting page
Error:  (https://bugreport.java.com/) after checking the Bug Database (https://bugs.java.com/)
Error:  for duplicates. Include error messages and the following diagnostic in your report. Thank you.
Error:  java.lang.NullPointerException: Cannot invoke "com.sun.source.util.DocTreePath.getTreePath()" because "path" is null
```

but with Java 20 all is green

ottlinger:
@slachiewicz the error seems to be related to https://bugs.openjdk.org/browse/JDK-8270866
Not sure if a newer JDK17.x may help.

PR #142, opened 2023-05-01
RAT-317: Maven plugin is too chatty

garydgregory:
Use DEBUG instead of INFO for some logging events


ottlinger:
Thanks for the PR - I'm going to add the changelog entry after merging the PR. Cheers

PR #149, opened 2023-07-24
RAT-251: Update changes.xml about ability for SPDX processing

Claudenw:
Updated changes.xml for pull request #148


ottlinger (changes requested):
Sorry I do not want to seem picky, but shouldn't it be:

dev="jochenw" due-to="Claude Warren"

as Claude is (not yet) a member of Creadur?

ottlinger (approved):
Thanks

PR #152, opened 2023-08-08
RAT-251: Resolve slow down issues from pull request #148

Claudenw:
#148 significantly slowed processing in cases with files that have no recognized licences.  This change significantly speeds up the SPDX processing without sacrificing or changing the functionality.


ottlinger:
Thanks @Claudenw ! I guess we should add more info to the webpage about how to use the new SPDX functionality. Thanks and good that the feature was tested "in real life".

PR #158, opened 2023-10-22
RAT-321: Fix for bad reporting of license status for non standard documents 

Claudenw:
Fix for bad reporting of license status for non standard documents (e.g. binary) as noted in RAT-321

Collected license filtering into new LicenseSetFactory class and applied it where expected. Added more tests to DefaultsTests
Added more tests to ReportConfigurtionTest


ottlinger:
Thanks for the fix - I'll apply it and verify.

ottlinger:
@Claudenw there seems to be a "breaking" change. ASL2-licensed files are marked as "ASL" in the report, although the report says:
`Files with Apache License headers will be marked AL`

Claudenw:
ASL licenses have copyright statements in them.  I don't know who uses that difference but it was in the original code base.  the name says that it is an "Applied Apache License"

Claudenw:
In the earlier version the Applied license would only trigger if the copyright came after something that identified the Apache license.  So the copyright had to be first.  Because it exited when the match was found the AL was registered for ASL as well.

I think we should remove ASL and the `not(copyright)` check from the default definition.

ottlinger:
As https://github.com/ottlinger/fotorenamer/pull/317 is a AL-licensed project I assume this would make sense to not break most of the existing RAT customers.

Claudenw:
OK.  So Licenses really don't have a unique category.  AL and ASL should both report as AL.  I can fix this fairly easily I think.

ottlinger:
I do not really now the exact difference between the two licenses but cannot find any official documentation about it - https://www.apache.org/licenses/LICENSE-2.0.html should be marked as AL.
Thanks for your help!
I'd prefer to break any users of applied ASL instead of regular AL one's ;)

ottlinger:
@Claudenw please use RAT-321 as prefix in your commits to properly get the commits accounted on the Jira ticket. Thanks.

PR #160, opened 2023-10-30
RAT-322: add option to be able to scan hidden directories and minor cleanups

jbonofre:
* add support to scan hidden directories flag to `DirectoryWalker`
* add scan hidden directories to the `ReportConfiguration`
* cleanup `Walker` moving `isRestricted` to the `DirectoryWalker` as it's specific
* expose scan hidden directories configuration in CLI, Ant task and Maven plugin

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
@@ -59,6 +59,7 @@ public class ReportConfiguration {
     private IOSupplier<InputStream> styleSheet = null;
     private IReportable reportable = null;
     private FilenameFilter inputFileFilter = null;
+    private boolean scanHiddenDirectories = false;
 
     /**
      * @return The filename filter for the potential input files.
@@ -194,7 +195,7 @@ public class ReportConfiguration {
      * Adds multiple licenses to the list of licenses. Does not add the licenses to
      * the list of approved licenses.
      *
-     * @param license The license t oadd.
+     * @param licenses The licenses to add.
      */
     public void addLicenses(Collection<ILicense> licenses) {
         this.licenses.addAll(licenses);
@@ -277,7 +278,6 @@ public class ReportConfiguration {
      * This value is ignored, if no license headers are added.
      *
      * @param copyrightMessage message to set.
-     * @see #setAddingLicenses(boolean)
      */
     public void setCopyrightMessage(String copyrightMessage) {
         this.copyrightMessage = copyrightMessage;
@@ -308,7 +308,6 @@ public class ReportConfiguration {
      * 
      * @param addLicenseHeaders enables/disables or forces adding of licenses
      * headers.
-     * @see #setAddingLicensesForced(boolean)
      * @see #setCopyrightMessage(String)
      */
     public void setAddLicenseHeaders(AddLicenseHeaders addLicenseHeaders) {
@@ -367,9 +366,26 @@ public class ReportConfiguration {
     }
 
     /**
+     *
+     * @return true if rat scan hidden directories (e.g. directory starting with dot), false to ignore hidden directories
+     */
+    public boolean isScanHiddenDirectories() {
+        return this.scanHiddenDirectories;
+    }
+
+    /**
+     * Define if rat should scan hidden directories (e.g. directory starting with dot) or not.
+     *
+     * @param scanHiddenDirectories true to scan hidden directories, false else.
+     */
+    public void setScanHiddenDirectories(boolean scanHiddenDirectories) {
+        this.scanHiddenDirectories = scanHiddenDirectories;
+    }
+
+    /**
      * Validates that the configuration is valid.
      * 
-     * @param a String consumer to log warning messages to.
+     * @param logger String consumer to log warning messages to.
      */
     public void validate(Consumer<String> logger) {
         if (reportable == null) {
--- a/apache-rat-core/src/main/java/org/apache/rat/walker/DirectoryWalker.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/walker/DirectoryWalker.java
@@ -37,8 +37,10 @@ public class DirectoryWalker extends Walker implements IReportable {
 
     protected static final FileNameComparator COMPARATOR = new FileNameComparator();
 
-    public DirectoryWalker(File file) {
-        this(file, (FilenameFilter) null);
+    private boolean walkHiddenDirectories = false;
+
+    public DirectoryWalker(File file, boolean walkHiddenDirectories) {
+        this(file, (FilenameFilter) null, walkHiddenDirectories);
     }
 
     /**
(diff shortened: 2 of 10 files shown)


jbonofre:
@ottlinger As discussed in Jira 😄

ottlinger:
@Claudenw is this in sync with what you proposed in RAT-323?

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -194,7 +195,7 @@ public void addLicense(ILicense license) {
      * Adds multiple licenses to the list of licenses. Does not add the licenses to
      * the list of approved licenses.
      *
-     * @param license The license t oadd.
+     * @param licenses The licenses to odd.
```
ottlinger, line 198: to add

jbonofre: Fixed

jbonofre:
I quickly discussed with @Claudenw on Slack. I think it's aligned.
As I detected issue in missing ASF headers in files located in hidden directories, it would be great to release a new rat version including this improvement.

ottlinger:
@jbonofre can you add an entry in the changelog? Thanks

jbonofre:
@ottlinger sure, I will :)

apache-rat-core/src/main/java/org/apache/rat/walker/DirectoryWalker.java
Claudenw, line 1: I think that the DirectoryWalker should accept a second FilenameFIlter rather than a boolean to limit the files, similar to the way the files are filtered.  This will give us more control over what files to read -- I see this as being necessary for some of the harmonization work that is coming.  Specifically reading specific hidden directories.

This change means that the configuration needs a pair of methods to set and get the input directory filter.

Claudenw (changes requested):

Claudenw:
Thank you for the DirectoryWalkerTest.

apache-rat-core/src/main/java/org/apache/rat/walker/DirectoryWalker.java
jbonofre: I'm not sure: I think it's already covered by the excludes. If we are able to read *all* directories (including the hidden files), then we can filter with exclude.
I think introducing a filter for "include" would be confusing. I would rather read all directories and delegate to exclude.
A valid proposal would be to scan all directories by default (including hidden ones) and use excludes.
Thoughts ?

apache-rat-plugin/src/main/java/org/apache/rat/mp/RatCheckMojo.java
```diff
@@ -109,6 +118,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             return;
         }
         ReportConfiguration config = getConfiguration();
+        config.setScanHiddenDirectories(scanHiddenDirectories);
```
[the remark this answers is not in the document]
jbonofre: Maybe I don't understand: `setScanHiddenDirectories()` is in `getConfiguration()` right ?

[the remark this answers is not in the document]
Claudenw: At line 175 in this file there is the method getConfiguration().  That method should take all the options from the UI and create  a ReportConfiguration that represents the state of the UI options.  line 121 should be moved into that method.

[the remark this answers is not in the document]
jbonofre: Ack, I do the change. Thanks.

apache-rat-core/src/main/java/org/apache/rat/walker/DirectoryWalker.java
Claudenw: I agree. However, if we switch to org.apache.commons.io.filefilter.AbstractFileFilter we can merge multiple filters together.

 What we can do is modify the ReportConfiguration so that

- by default set the directory filter to ignore dot prefixed directories.
- if addDirectoryFilter() is called with NULL we remove the directory filter.
- If addDirectoryFilter() is called with a non-null argument then the argument is "and"ed to the filter.

Alternatively we could create a FilterBuilder that constructs filters using "and" and "or" methods to build an AbstractFileFilter and accept the builder as the argument for a 'setDirectoryFilter()' method.  Then passing NULL could set the internal value to org.apache.commons.io.filefilter.TrueFileFilter (accepting all directories).

I think that creating a builder will make specifying options in ANT UI easier.  However, for now we can make due with adding a directoryFilter instance variable of type org.apache.commons.io.filefilter.AbstractFileFilter  and setting it to ignore dot prefixed files by default.  Then your current method can simpley change that to org.apache.commons.io.filefilter.TrueFileFilter.  Later we can figure out how to specify more complex filters.

Claudenw: If you add the AbstractFileFilter then it is probably advisable to change the internal representation of ReportConfiguration.inputFileFilter to that type as well.

jbonofre: OK, let me update the PR then. Thanks.

jbonofre: I experimented using `IOFileFilter` (leveraging `HiddenFileFilter`, `FalseFileFilter`, etc).
Adding `addDirectoryFilter()` to `ReportConfiguration` is OK (and using `IOFileFIlter#and()` to add multiple filters.

At code level, it works fine, but I have a concern about the "usage", especially on the CLI and Maven. I think for the CLI, we should keep the proposed option (`--scan-hidden-directories` for CLI for instance) to simplify the use.

Thoughts ?

jbonofre: I opened #166 using `DirectoryFilter`. @Claudenw @ottlinger let me know which approach you prefer :)

Claudenw: I think my approach would be adding the `addDirectoryFilter` to the `ReportConfiguration` so we have the groundwork for the other cases.

Then add the `--scan-hidden-directories` option (or equivalent) to all the clients. (CLI, Maven, ANT).  Making sure the --scan-hidden-directories is implemented across all the clients is part of RAT-323.

jbonofre: Agree, it's what I did in the other PR. I can cancel this one in favor of the new one. Can you please take a look on the new PR ?

ottlinger:
@jbonofre can we close this MR here as discussions are continuing in https://github.com/apache/creadur-rat/pull/166 (just to not confuse any reviewer/reader). Thanks for your contributions.

jbonofre:
@ottlinger absolutely :)

PR #161, opened 2023-10-30
RAT-320: Add -o/--output option to CLI

jbonofre:


apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -349,29 +357,20 @@ private Report() {
...
     private static IReportable getDirectory(String baseDirectory, ReportConfiguration config) {
-        try (PrintStream out = new PrintStream(config.getOutput().get())) {
-            File base = new File(baseDirectory);
-            if (!base.exists()) {
-                out.print("ERROR: ");
-                out.print(baseDirectory);
-                out.print(" does not exist.\n");
-                return null;
-            }
+        File base = new File(baseDirectory);
```
jbonofre, line 360: FYI, `getDirectory()` method doesn't cleanly close the `Writer` from config, e.g. we have a bug: nothing is printed to `System.out`.
This change here fix the issue.

ottlinger:
Would it be possible to add a test for the newly created behaviour? (apart from the changelog entry ;))

I guess that the docs need to be updated, too:
* https://github.com/apache/creadur-rat/blob/master/apache-rat/src/site/apt/index.apt.vm
* https://github.com/apache/creadur-rat/blob/master/apache-rat/README-CLI.txt

jbonofre:
@ottlinger good points. Let me update the PR accordingly. Thanks.

apache-rat-core/src/main/java/org/apache/rat/Report.java
Claudenw: In the new code base (after RAT-321) the ReportConfiguraiton.out is now a org.apache.commons.io.function.IOSupplier so the output does not get constructed until it is used and then it is done in the try with resource closable.

You should be able to rebase and have the "o" option create the IOSupplier.

jbonofre: Catcha. Let me rebase and update accordingly.

jbonofre:
I will rebase and update accordingly.

apache-rat-core/src/main/java/org/apache/rat/Report.java
jbonofre: @Claudenw is it already merged on `master` ? I tested on `master` and, even if it should with `out` is `null`, and nothing is printed on `System.out` using CLI. If I remove the lines in this change, it works. So I'm pretty sure that this part is wrong and the `Writer` is not closed cleanly.

For instance:

```
$ java -jar apache-rat-0.16-SNAPSHOT.jar /Users/jbonofre/mytest
$
```

So nothing is printed on `System.out` whereas it shoul.

It's a bug for me, fixed with my change.

Claudenw: I see that what you have done is remove the output from the method and let the exceptions provide the info.

apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -349,29 +357,20 @@ private Report() {
      * @return the IReportale instance containing the files.
      */
     private static IReportable getDirectory(String baseDirectory, ReportConfiguration config) {
-        try (PrintStream out = new PrintStream(config.getOutput().get())) {
```
Claudenw, line 352: This line should close the output.

Claudenw (approved):
I see that your code is cleaner.  It does raise a question about the concept of using an IOSupplier but that seems to work elsewhere.  I think the issue with the original code was that it opened the IOSupplied output and then closed it when it exited, but the code that detected the error did not throw an exception so at a later stage the IOSupplier was opened again and stepped on the original writing. 

As I said your solution is cleaner so let's use it.  However, we should be on the lookout for other issues where we open and close the IOSupplier multiple times.

apache-rat-core/src/main/java/org/apache/rat/Report.java
ottlinger: @jbonofre I merged other changes concerning RAT-321. Pls retry.

ottlinger:
Please make sure to rebase as changes came in via https://github.com/apache/creadur-rat/pull/163/files - thanks.

jbonofre:
I will but I need time : we have been hit by the storm and no electricity for now :/

Claudenw:
I have a re based branch of JB's work.  If you want I can submit that.

jbonofre:
I should be back online today. So I will tackle that. Thanks !

jbonofre:
I'm back :) I'm rebasing and fixing conflict. I will update the PR very soon.

jbonofre:
Back on Rat 😄

apache-rat-core/src/main/java/org/apache/rat/Report.java
jbonofre: I confirm that it works now on `master`. I will rebase and clean this PR accordingly.

jbonofre:
Rebase done, I'm adding tests now as requested by @ottlinger

Claudenw:
Looks very good to me (LVGTM?) .  I await the tests with baited breath -- No I did not eat the worms.

jbonofre:
@Claudenw @ottlinger I added a couple of simple tests.

jbonofre:
@ottlinger I think we are good about this one. Can you please take a look ?

apache-rat/src/site/apt/index.apt.vm
```diff
@@ -75,17 +75,17 @@ java -jar apache-rat/target/apache-rat-${project.version}.jar
 
 
 +------------------------------------------+
-usage: java -jar apache-rat/target/apache-rat-${project.version}.jar
- [options] [DIR|TARBALL]
+usage: java -jar apache-rat/target/apache-rat-CURRENT-VERSION.jar
```
ottlinger, line 78: I've asked the same question in the other PR - could we stick with project.version here in order to generate the correct version into the file?

ottlinger:
I'll add the changelog entry upon merging - thanks again for your contribution @jbonofre

jbonofre:
@ottlinger should I update `RELEASE_NOTES.txt` ? I thought it's updated just at release time.

apache-rat/src/site/apt/index.apt.vm
jbonofre: Fixed

jbonofre: FYI, I've updated site/README with all missing options (copy/paste of the `--help` output).

ottlinger:
> @ottlinger should I update `RELEASE_NOTES.txt` ? I thought it's updated just at release time.

The txt file is generated, while the actual changelog is src/changes/changes.xml. I'll take care of that.

ottlinger:
Thanks again for this contribution @jbonofre

PR #163, opened 2023-11-02
RAT-328 Fix for IOSupplier issue includes javadoc updates

Claudenw:
Ensured the System.out does not get closed and updated javadocs. 

Added class to allow others to wrap similar streams in a no close format.


ottlinger:
Thanks for the fix, I'll add a changelog entry.

PR #165, opened 2023-11-12
RAT-326: Fix javadoc build errors

ottlinger:


Claudenw:
should we take this is small chunks?  I think we should merge these changes but leave RAT-236 open until we get a build that has no javadoc errors.  This way we can review smaller tranches of files rather than trying to do everything at once.

ottlinger:
First batch of changes merged - we seem to run into https://bugs.openjdk.org/browse/JDK-8270866 when I run 
mvn javadoc:javadoc inside of apache-rat-core.

PR #168, opened 2023-11-25
added architecture.md and javadocs for RatReport

Claudenw:
No jira for this change.  I added a simple overview of the architecture in the hopes that we will expand it and keep it up to date so that new developers will have some idea of how this works.


ottlinger:
Thanks @Claudenw, this is a perfect starter for future contributions!

PR #175, opened 2023-12-08
RAT-338: Bulk refresh for plugin

cstamas:
Refresh deps but also code. And please release, as current last release is a Maven 2 plugin:
https://deps.dev/maven/org.apache.rat%3Aapache-rat-plugin/0.15/dependencies

---

https://issues.apache.org/jira/browse/RAT-338

### DIFF
--- a/apache-rat-plugin/src/main/java/org/apache/rat/mp/AbstractRatMojo.java
+++ b/apache-rat-plugin/src/main/java/org/apache/rat/mp/AbstractRatMojo.java
@@ -26,7 +26,6 @@ import static org.apache.rat.mp.util.ExclusionHelper.addPlexusAndScmDefaults;
 import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -34,6 +33,7 @@ import java.io.Reader;
 import java.lang.reflect.UndeclaredThrowableException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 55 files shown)


ottlinger:
@cstamas thanks for your contributions - we are about to push out a new release, but work is still going on concerning some performance regressions apart from moving to Maven3.
Pls rebase your PR as some of the dependabot updates were merged in.

Can you be so kind to create a https://issues.apache.org/jira/browse/RAT an reference it in your PR? Thanks again for your help!

pom.xml
```diff
@@ -52,12 +52,13 @@ agnostic home for software distribution comprehension and audit tools.
     <maven.compiler.source>${javaVersion}</maven.compiler.source>
     <maven.compiler.target>${javaVersion}</maven.compiler.target>
     <!-- This is the version of Maven required to use the Rat Maven Plugin -->
-    <mavenVersion>3.2.5</mavenVersion>
+    <mavenMinVersion>3.2.5</mavenMinVersion>
+    <mavenVersion>3.9.6</mavenVersion>
```
ottlinger, line 56: fyi: I started a discussion on the mailing list concerning this update ...... let's see if there are any complaints about that one.

cstamas: The point of this update: "we" (Maven project) did it wrong, compile against latest (3.9.6) but allow runtime 3.2.5 (min maven). The reason to do so is to pick up deprecations and get notified what is changing, but source and binary compatibility guarantees it to run on min version as well. If you compile against 3.2.5 forever, you get totally disconnected about latest changes and deprecations in Maven APIs.

apache-rat-plugin/src/main/java/org/apache/rat/mp/AbstractRatMojo.java
```diff
@@ -392,7 +392,7 @@ private void setIncludes(DirectoryScanner ds) throws MojoExecutionException {
                 }
                 includeList.addAll(getPatternsFromFile(f, charset));
             }
-            ds.setIncludes(includeList.toArray(new String[includeList.size()]));
+            ds.setIncludes(includeList.toArray(new String[0]));
```
ottlinger, line 395: Following https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#toArray-T:A- setting the array's size prevents creating a new one.

pom.xml
cstamas: Also, this affects _compilation only_, as at runtime, whatever Maven version you use, will provide the runtime stuff, and this does not mean Maven 3.9.6 is must or needed or whatever. Again, this is only to get compiler emit deprecation messages to notify devs (like use of deprecated type ArtifactRepository in one mojo is).

ottlinger: Thanks for this context. So basically it's the same as https://github.com/apache/creadur-rat/pull/172 concerning the maven deps?

cstamas: Not quite: that is "machine" going for highest, for example it left unreal "min" version in place:
https://github.com/apache/creadur-rat/blob/b0442fd49026b1929e9a88f2ea1c655dca106115/pom.xml#L469

Maven 3.0.x should be avoided (due package rename happened between 3.0.x and 3.1.x). We in Maven also set "lowest version" to 3.2.5 these days, it covers from 2014 until today, I hope 10 years is good enough.

ottlinger:
@cstamas thanks for your contributions, fixes and explanations.

PR #189, opened 2024-01-10
RAT-343: reimplement old config elements for custom licenses in Maven plugin configurations

Claudenw:
fix for RAT-343

Reinstated some removed classes and added Deprecaton annotations.
Added some methods in AbstractRatMojo and marked them for Deprecation so that we will remember to remove them and refactor when we get to version 1.0.

This change does not yet include the updates for the documentation.


ottlinger:
Thanks for the update. I'll see if I can easily change the documentation as it contains the new way to define custom licenses.

PR #192, opened 2024-01-12
RAT-325 : Added check to skip expensive processing in SPDX

Claudenw:
In my small test case this changed the timing from 55 sec to 3 sec.

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/Report.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/Report.java
@@ -126,13 +126,12 @@ public class Report {
      */
     public static void main(String[] args) throws Exception {
         Options opts = buildOptions();
-
         CommandLine cl;
         try {
             cl = new DefaultParser().parse(opts, args);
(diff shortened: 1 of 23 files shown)


ottlinger:
Can you push a new commit? Strangely the build runs locally. Maybe GHA just needs a 2nd chance ;)

ottlinger:
My local builds run ok on an Ubuntu, but GHA fails .... strange.

ottlinger:
As I merged in the changes related to RAT-346; pls rebase this branch and let us tackle the problems with the tests then. Thanks again for your help & efforts!

ottlinger:
The changes look promosing; I will merge and we will try on the openmeetings project. Thanks again.

ottlinger:
@Claudenw I ran the current SNAPSHOT on openmeetings and realized that all processed files are printed out on level INFO. I think we should change this to DEBUG as people complained about log spam of RAT runs in previous releases. Would you mind opening a new PR for that change? Thanks

PR #193, opened 2024-01-12
RAT-348: Update gitignore reader

nielsbasjes:
Just updating to the latest version.
This fixes some of the special expressions in a gitignore file that are allowed yet poorly documented.


ottlinger:
Thanks for the update; will add a changelog entry accordingly.

PR #194, opened 2024-01-13
RAT-343: Add test to validate allowing a custom license

nielsbasjes:
A simple test to ensure custom licenses work

### DIFF
--- a/apache-rat-plugin/src/main/java/org/apache/rat/mp/Family.java
+++ b/apache-rat-plugin/src/main/java/org/apache/rat/mp/Family.java
@@ -38,4 +38,8 @@ public class Family {
         return builder.setLicenseFamilyCategory(id).setLicenseFamilyName(name).build();
     }
 
+    @Override
+    public String toString() {
+        return '{' + id + ':' + name + '}';
+    }
 }
--- a/apache-rat-plugin/src/main/java/org/apache/rat/mp/License.java
+++ b/apache-rat-plugin/src/main/java/org/apache/rat/mp/License.java
@@ -57,4 +57,9 @@ public class License extends EnclosingMatcher {
                 .setName(name).setNotes(notes).build(context);
     }
 
+    @Override
+    public String toString() {
+        return '{' + family + ':' + id + ':' + name + '}';
+    }
+
 }
--- a/apache-rat-plugin/pom.xml
+++ b/apache-rat-plugin/pom.xml
@@ -94,6 +94,8 @@
               <exclude>src/test/resources/unit/RAT-335-GitIgnore/**/*</exclude>
               <exclude>**/*.iml</exclude>
               <exclude>src/test/resources/unit/RAT-343/*</exclude>
+              <!-- These files have a custom license header (non Apache) to test validating custom licenses -->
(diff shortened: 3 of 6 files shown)


nielsbasjes:
To make the test good I needed to remove the Apache license ... which triggered an NPE fixed in this separate merge request: https://github.com/apache/creadur-rat/pull/196

nielsbasjes:
So there IS an NPE in the integration test but for some reason I do not know yet the build does not fail.

```
[INFO] --- invoker:3.6.0:integration-test (integration-test) @ apache-rat-plugin ---
[INFO] Building: RAT-268/pom.xml
[INFO]           RAT-268/pom.xml .................................. SUCCESS (3.393 s)
[INFO] Building: it1/pom.xml
[INFO] run post-build script verify.groovy
[INFO]           it1/pom.xml ...................................... SUCCESS (3.249 s)
[INFO] Building: CustomLicense/pom.xml
[INFO]   The build exited with code 1. See /home/runner/work/creadur-rat/creadur-rat/apache-rat-plugin/target/invoker/CustomLicense/build.log for details.
[INFO]           CustomLicense/pom.xml ............................ FAILED (2.081 s)
[INFO] Building: it4_RAT-168/pom.xml
[INFO] run post-build script verify.groovy
[INFO]           it4_RAT-168/pom.xml .............................. SUCCESS (2.848 s)
```

apache-rat-plugin/src/main/java/org/apache/rat/mp/Family.java
```diff
@@ -38,4 +38,8 @@ public ILicenseFamily build() {
         return builder.setLicenseFamilyCategory(id).setLicenseFamilyName(name).build();
     }
 
+    @Override
+    public String toString() {
+        return '{' + id + ':' + name + '}';
+    }
```
Claudenw, line 44: Why do you need a toString() on the builder?  I don't see any uses of it in your code.

apache-rat-plugin/src/main/java/org/apache/rat/mp/License.java
```diff
@@ -57,4 +57,9 @@ public ILicense build(SortedSet<ILicenseFamily> context) {
                 .setName(name).setNotes(notes).build(context);
     }
 
+    @Override
+    public String toString() {
+        return '{' + family + ':' + id + ':' + name + '}';
+    }
+
```
Claudenw, line 64: Why do you need a toString() on the builder?  I don't see any uses of it in your code.

apache-rat-plugin/src/main/java/org/apache/rat/mp/Family.java
nielsbasjes: Correct. I needed this to read the debug output when running mvn -X. Without this it shows the class name and the hash... which makes it useless output.

apache-rat-plugin/src/main/java/org/apache/rat/mp/License.java
nielsbasjes: Yes, same as the other one.

ottlinger:
Thanks for the test and working example.

PR #195, opened 2024-01-13
RAT-346: warn on duplicate license family

Claudenw:
Added code to warn on duplicate license or license family.

Also upgraded to Junit 5 as it was needed for the testing.


apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -269,29 +264,41 @@ static FilenameFilter parseExclusions(List<String> excludes) {
...
-
-        Option out = new Option("o", "out", true, "Define the output file where to write report (default is System.out)");
+        String licFilterValues = String.join(", ", 
+                Arrays.stream(LicenseFilter.values()).map(LicenseFilter::name).collect(Collectors.toList()));
+
+        Options opts = new Options()
+        
+        .addOption(
+                Option.builder().hasArg(true).longOpt(LIST_FAMILIES)
+                .desc("List the defined license families Valid options are: "+licFilterValues+". Default=none")
```
ottlinger, line 274: List the defined license families (default is none). Valid options are: .....

apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -269,29 +264,41 @@ static FilenameFilter parseExclusions(List<String> excludes) {
...
+                .build())
+        .addOption(
+                Option.builder().hasArg(true).longOpt(LIST_LICENSES)
+                .desc("List the defined licenses Valid options are: "+licFilterValues+". Default=none")
```
ottlinger, line 278: List the defined licenses (default is none). Valid options are: .....

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -114,16 +174,18 @@ public void setReportable(IReportable reportable) {
...
         return styleSheet;
     }
 
     /**
-     * Sets the style sheet for custom processing.
-     * The IOSupplier may be called multiple times, so the input stream must be 
-     * able to be opened and closed multiple times.
+     * Sets the style sheet for custom processing. The IOSupplier may be called
+     * multiple times, so the input stream must be able to be opened and closed
+     * multiple times.
```
ottlinger, line 187: btw: I guess that this new behaviour is the cause of https://issues.apache.org/jira/browse/RAT-344 ;)

apache-rat-core/src/test/java/org/apache/rat/analysis/matchers/CopyrightMatcherTest.java
```diff
@@ -53,91 +53,99 @@ public class CopyrightMatcherTest {
...
+          Arrays.stream(args).map(origin -> new String[] { pfx + origin[0], pfx + origin[1] })
+                  .forEach(arry::add);
         }
-        result[5] = arry.toArray(new String[arry.size()][2]);
-        return result;
+        return arry.toArray(new String[arry.size()][2]);
     }
 
+            
+//    private static Arguments expandResults(String name, String startDate, String endDate, String owner,
```
ottlinger, line 83: pls remove old methods as commented-out code will never be removed again ... thx

apache-rat-core/src/test/java/org/apache/rat/analysis/matchers/CopyrightMatcherTest.java
```diff
@@ -53,91 +53,99 @@ public class CopyrightMatcherTest {
...
 
-    public CopyrightMatcherTest(String testName, String start, String stop, String owner, String[][] pass,
+//    public CopyrightMatcherTest(String testName, String start, String stop, String owner, String[][] pass,
+//            String[][] fail) {
+//        verify(testName, pass, fail);
+//        matcher = new CopyrightMatcher(start, stop, owner);
+//        this.pass = pass;
+//        this.fail = fail;
+//        this.testName = testName;
+//    }
```
ottlinger, line 121: pls remove old methods as commented-out code will never be removed again ... thx

apache-rat-core/src/test/java/org/apache/rat/annotation/TestLicenseAppender.java
```diff
@@ -18,25 +18,36 @@
...
+    
+    @BeforeAll
+    public static void setup() throws IOException {
+        baseTempFolder = Files.createTempDirectory("tla").toFile();
+    }
+    
+    @AfterAll
+    public static void teardown() throws IOException {
+        FileUtils.deleteDirectory(baseTempFolder);
+    }
```
ottlinger, line 50: I'd prefer to use @TempDir here ..... as suggested in
https://www.baeldung.com/junit-5-temporary-directory
(we had trouble with /tmp running out of space with hand-crafted solutions in case of test failures a.s.o)

apache-rat-core/src/test/java/org/apache/rat/walker/DirectoryWalkerTest.java
```diff
@@ -25,28 +25,41 @@
...
+    
+    @BeforeEach
+    public void setup() throws IOException {
+        toWalk = Files.createTempDirectory("dwt").toFile();
+    }
+    
+    @AfterEach
+    public void teardown() throws IOException {
+        FileUtils.deleteDirectory(toWalk);
+    }
```
ottlinger, line 58: I'd prefer @TempDir here as well .....

apache-rat-plugin/src/test/java/org/apache/rat/mp/util/ScmIgnoreParserTest.java
```diff
@@ -86,13 +84,19 @@ public void parseFromTargetDirectoryHopefullyWithoutSCMIgnores() {
...
+            ignore = Files.createTempFile("sip", null).toFile();
+            assertTrue(ignore.exists());
+            writeToFile(IGNORE_EXAMPLE, ignore);
+        
+            final List<String> excludes = getExcludesFromFile(log, ignore);
+            assertFalse(excludes.isEmpty());
+            assertEquals(2, excludes.size());
+        } finally {
+            ignore.delete();
+        }
```
ottlinger, line 98: More @TempDir ......

ottlinger:
@Claudenw not sure why you removed the API module ..... if I understood it correctly we want to harmonize things among all the available RAT-options (e.g. Ant,Maven,CLI). I'd assume an API module to be handy in that regard.

ottlinger:
@jbonofre just fyi - a great deal of changes will come in with the jUnit5 update.

jbonofre:
@ottlinger thanks for the heads up. I will be back on rat work tomorrow.

Claudenw:
@ottlinger thx for the quick review

I'll make the change to the @TempDir annotation -- I think there are 2 files where it is needed.

I removed the API module because it consisted of a) class files that were not being used in the code but that were misleading in name; and more importantly b) because there were a number of test cases that would have been converted to Junit5 only to be removed again later.

The harmonization is not necessarily the use of a single API module but rather that the options in each of the UIs should be similarly constructed (same basic phrasing)  not have 2 UIs where one says "--enable-default-licenses" (default off) and the other says "--disable-default-licenses" (default on).  Also there are options that have been added to one UI that are general in nature (i.e. not part of managing the specific platform the UI is on) and are not available in the other UIs.

I will remove the commented out code.  I should have before I submitted this pull request, but skipped it in favor of getting this code change infront of more eyes quickly.

I also have to make a pass to update javadocs.

@jbonofre  Most of the changes in this request are the migration to Junit 5.  The significant changes are the creation of the classes in the o.a.r..utils directory, the addition of methods to set the associated properties in  ReportConfiguration along with the tests for those classes and changes.

The request was to warn on duplication so the solution submitted does the following:

- Creates a ReportingSet class that is a SortedSet implementation (wrapping and delegating to a SortedSet).  it intercepts the calls to `add` and `addAll` to determine if the new elements are already in the set and take an action.

    - If logging is enabled (not set to None on the very thing logging wrapper `Log`) write a message about the collision to the log at the specific level.
    - If the internal action (ReportingSet.Option) is set to `IGNORE` the change is ignored, `OVERWRITE` the old entry is replaced, `FAIL` an exception is thrown.

Finally the ReportingSet has an instance variable called `duplicateFmt` that is a `Function<T,String>` that produces the information to display in the message.

The change to ReportConfiguration are to use ReportingSet to contain the Licenses and LicenseFamilies, and supporting methods to configure (as noted above).

Claudenw:
@ottlinger Thanks for taking the time to clean this up.  It looks good to me.

ottlinger:
@Claudenw pls give a ping if you have any local changes that need to be pushed. I`d love to merge this PR in order to continue analysing the problems on https://github.com/apache/creadur-rat/pull/192

ottlinger:
Thanks @Claudenw for the simplifications and the migration efforts!

PR #196, opened 2024-01-13
RAT-349: Fix NPE on addDefaultLicenses=false

nielsbasjes:
The place where the NPE occurs is here 
https://github.com/apache/creadur-rat/blob/21b726a0945f905d67ef8854950aa1e4530873a3/apache-rat-core/src/main/java/org/apache/rat/Reporter.java#L56

Because no config is set the stylesheet is null which NPEs on the `.get()`.

If you normally set a config without a styleSheet something similar is done.

### DIFF
--- a/apache-rat-plugin/src/main/java/org/apache/rat/mp/AbstractRatMojo.java
+++ b/apache-rat-plugin/src/main/java/org/apache/rat/mp/AbstractRatMojo.java
@@ -318,6 +318,8 @@ public abstract class AbstractRatMojo extends AbstractMojo {
         reportDeprecatedProcessing();
         if (addDefaultLicenses) {
             config.setFrom(getDefaultsBuilder().build());
+        } else {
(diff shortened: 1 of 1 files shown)


Claudenw:
@nielsbasjes  thanks for the fix, but I am concerned that something else is wrong.  

If I run  RAT from the CLI and givie it the command line option `--no-default-licenses` it reports 
```
Exception in thread "main" org.apache.rat.ConfigurationException: You must specify at least one license
	at org.apache.rat.ReportConfiguration.validate(ReportConfiguration.java:495)
	at org.apache.rat.Report.main(Report.java:155)
```

How did you manage to get a null stylesheet while still having `isStyleReport()` return `true`?

nielsbasjes:
Check this because that is how I found this problem https://github.com/apache/creadur-rat/pull/194
I wanted the test to only allow my custom  license and it crashes with an NPE.

ottlinger:
Thanks for the finding.

PR #198, opened 2024-01-14
RAT-348: Update GitIgnore reader to version 1.3.1

nielsbasjes:
@ottlinger FYI


ottlinger:
Let us wait until the artifact is properly available in all Maven repos and the builds are green. Thanks!

PR #204, opened 2024-01-17
RAT-325: Changed CLI-default logging level to WARN

Claudenw:
Rather than moving the file lists to the noisier DEBUG level I update the default logging level to WARN.  This seems appropriate for CLI.

The default logging level only applies to the CLI and  tests.
In all other extant UIs the logging is performed by the native system and so the logging needs to be set there.

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/Report.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/Report.java
@@ -144,7 +144,7 @@ public class Report {
                 Log.Level level = Log.Level.valueOf(cl.getOptionValue(LOG_LEVEL).toUpperCase());
                 DefaultLog.INSTANCE.setLevel(level);
             } catch (IllegalArgumentException e) {
-                DefaultLog.INSTANCE.warn(String.format("Invalied Log Level (%s) specified.", cl.getOptionValue(LOG_LEVEL)));
+                DefaultLog.INSTANCE.warn(String.format("Invalid Log Level (%s) specified.", cl.getOptionValue(LOG_LEVEL)));
                 DefaultLog.INSTANCE.warn(String.format("Log level set at: %s", DefaultLog.INSTANCE.getLevel()));
             }
         }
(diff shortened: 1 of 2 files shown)


apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -137,6 +139,15 @@ public static void main(String[] args) throws Exception {
                     // for "cl"
         }
 
+        if (cl.hasOption(LOG_LEVEL)) {
+            try {
+                Log.Level level = Log.Level.valueOf(cl.getOptionValue(LOG_LEVEL).toUpperCase());
+                DefaultLog.INSTANCE.setLevel(level);
+            } catch (IllegalArgumentException e) {
+                DefaultLog.INSTANCE.warn(String.format("Invalied Log Level (%s) specified.", cl.getOptionValue(LOG_LEVEL)));
```
ottlinger, line 147: Invalid ;)

apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -327,6 +338,10 @@ static Options buildOptions() {
         Option dir = new Option("d", "dir", false, "Used to indicate source when using --exclude");
         opts.addOption(dir);
 
+        opts.addOption( Option.builder().argName("level").longOpt(LOG_LEVEL)
+                .hasArgs().desc("sets the log level.  Valid options are: DEBUG, INFO, WARN, ERROR, NONE")
```
ottlinger, line 342: Didn't we use OFF at other enums within the code instead of NONE?

apache-rat-core/src/main/java/org/apache/rat/Report.java
Claudenw: fixed

apache-rat-core/src/main/java/org/apache/rat/Report.java
Claudenw: Your are correct. I fixed it.  I really could use the new apache commons CLI but it is not out of test.  Helps with getting enums from the command line.

ottlinger:
Thanks again, I will add the changelog entry directly on master after merging.

PR #205, opened 2024-01-21
RAT-355: Add configuration dump to report

Claudenw:
Closes Rat-355.

Adds CLI options.

--list-licenses all|approved|none  (default: none)
--list-families all | approved | none (default: none)
--dry-run

if either of the --list-X  options is not "none" the reporter will add a "rat-config" section to the report XML document.  This section has the same structure as the XML configuration file.

if --dry-run is specified then no files are updated and no files are scanned.


apache-rat-core/src/main/java/org/apache/rat/inspector/AbstractInspector.java
```diff
@@ -0,0 +1,120 @@
...
+        return type;
+    }
+
+    @Override
+    public String getCommonName() {
+        return name;
+    }
+
+    public static Inspector license(ILicense license, Inspector matcher) {
+        return new AbstractInspector(Inspector.Type.License, "license") {
```
ottlinger, line 47: Would one want to extract all of these license-config-related constants into a class where we could easily add documentation about the semantics of each option/value/parameter?

apache-rat-core/src/main/resources/org/apache/rat/default.xml
```diff
@@ -158,7 +158,7 @@
 					may
 					be prepared copied published and distributed</text>
 				<copyright
-					owner="OASIS Open" start="[0-9]{4}" />
+					owner="OASIS Open"/>
```
ottlinger, line 161: What is the semantics of this change? Should there be a test for this OASIS license that needs to be changed here?

Claudenw: The start date specified is the default if no data is given.  So I removed it as generally we want a regular year in there, not the regex pattern.

apache-rat-core/src/main/java/org/apache/rat/inspector/AbstractInspector.java
Claudenw: The parameters for the XML output should all be defined in the documentation about the XML configuration file.  That is partly done.  I opened a ticket about it.  I have some questions about how to go about documenting  things that span the various UIs.

Claudenw: The problem (for reporting) is that the current system allows creation of new IHeaderMatcher types and adding those matchers to the system.  This means that the output has to be able to handle matchers that we don't know about yet.

I think the way to approach this might be to have each system class that represents an item in the have a `getDescription()` method that will return a Description that is basically the same as the Inspector, but is defined within the class.  This will handle the new class problem noted above, as well as provide a mechanism to construct documentation for all the options and/or display that information to the user.

Claudenw:
Closing in favor of #227 which has a better solution

PR #206, opened 2024-01-22
RAT-353, MPLUGIN-507: Enable maven-plugin-report-plugin to see NPE with Maven 4.x only

ottlinger:
Filed a bug in the maven-plugin-report-plugin: [MPLUGIN-507](https://issues.apache.org/jira/browse/MPLUGIN-507)

[RAT-353](https://issues.apache.org/jira/browse/RAT-353) brought to the light that the following files were not generated in RAT 0.16 and 0.16.1:
* plugin-info.html
* check-mojo.html
* rat-mojo.html
due to a breaking change in maven-plugin-annotations v3.11.0 that separated the generation logics into a new plugin a new configuration is needed. This new plugin yields a NPE and breaks the site build.

### DIFF
--- a/.buildtools/generateStagingSiteInWebpageRepo
+++ b/.buildtools/generateStagingSiteInWebpageRepo
@@ -13,5 +13,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-mvn clean site:site site:stage
+./mvnw clean site:site site:stage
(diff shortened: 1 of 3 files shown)


ottlinger:
NPE only happens with Maven 4.x;
adapted helper script for site generation to use Maven wrapper version to avoid these confusions in the future.

PR #224, opened 2024-03-18
RAT-77: output files with missing headers

Claudenw:
Fix for RAT-77

This was mostly fixed earlier.  This pull 

- adds another stylesheet to the package (missing-headers.xsl) to the two that already existed: plain-rat (default), missing-headers, and unapproved-licenses
- allows the -s option to use a short name (e.g. `-s missing-headers` or `-s unapproved-licenses`
- updates the documentation to talk about the stylesheets.

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/Report.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/Report.java
@@ -314,15 +314,13 @@ public class Report {
         opts.addOption(null, SCAN_HIDDEN_DIRECTORIES, false, "Scan hidden directories");
 
         OptionGroup addLicenseGroup = new OptionGroup();
-        String addLicenseDesc = "Add the default license header to any file with an unknown license that is not in the exclusion list. "
-                + "By default new files will be created with the license header, "
-                + "to force the modification of existing files use the --force option.";
-
         // RAT-85/RAT-203: Deprecated! added only for convenience and for backwards
         // compatibility
-        Option addLicence = new Option(ADD_OLD, "addLicence", false, addLicenseDesc);
+        Option addLicence = new Option(ADD_OLD, false, "(deprecated) Add the default license header to any file with an unknown license.  Use '-a' or ---addLicense instead.");
         addLicenseGroup.addOption(addLicence);
-        Option addLicense = new Option(ADD, "addLicense", false, addLicenseDesc);
+        Option addLicense = new Option(ADD, "addLicense", false, "Add the default license header to any file with an unknown license that is not in the exclusion list. "
+                + "By default new files will be created with the license header, "
+                + "to force the modification of existing files use the --force option.");
         addLicenseGroup.addOption(addLicense);
         opts.addOptionGroup(addLicenseGroup);
 
--- a/apache-rat-core/src/main/java/org/apache/rat/Report.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/Report.java
@@ -21,8 +21,10 @@ package org.apache.rat;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.PrintStream;
 import java.io.Serializable;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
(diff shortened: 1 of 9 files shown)


src/site/apt/index.apt.vm
```diff
@@ -111,6 +113,19 @@ java -jar apache-rat-${project.version}.jar --help
...
+ 
+ ** Styling output
+ 
+ Rat allows you to style the output as you see fit.  Three stylesheets are included in the rat package.
+ 
+   plain-rat  The default stylesheet.
+   
+   unapproved-licenses  Lists only the files with unapproved licenses.
+   
+   missing-headers  List only files that are missing headers.
```
Claudenw, line 125: @ottlinger  I don't know how to make this work.  But I wanted bullet points for these lines,  However, the build failed when I tried to match what was above in the "How do I use RAT" section.

src/site/apt/index.apt.vm
```diff
@@ -111,6 +113,17 @@ java -jar apache-rat-${project.version}.jar --help
 +------------------------------------------+
 
  Read more {{{./apache-rat-plugin/index.html} here}}.
+ 
+ ** Styling output
+ 
+ Rat allows you to style the output as you see fit.  Three stylesheets are included in the rat package.
+ 
+    plain-rat  The default stylesheet.
```
ottlinger, line 121: Wouldn't it make sense to link to the stylesheets from the webpage? So people could directly take a look?

Claudenw: It might.  Do you have a fix for that?  Also, I couldn't get the APT interpreter to build the documentation.  I don't know what I did wrong.  Any pointers would be appreciated.

Claudenw:
@ottlinger Should be fixed now.

apache-rat/src/site/apt/index.apt.vm
```diff
@@ -79,12 +79,10 @@ usage: java -jar apache-rat/target/apache-rat-${project.version}.jar
...
 Available options
- -a,--addLicence                Add the default license header to any file
-                                with an unknown license that is not in the
-                                exclusion list. By default new files will
-                                be created with the license header, to
-                                force the modification of existing files
-                                use the --force option.
+ -a                             (deprecated) Add the default license
+                                header to any file with an unknown
+                                license.  Use '-a' or ---addLicense
```
ottlinger, line 84: -A or --addLicense ?!

Claudenw: good catch.  Fixed

ottlinger: Did you push the branch as Github is still showing the typo?

Claudenw: There must have been a github issue.  I checked this AM and I had pushed and currently github does show the changes in this pull request.  Please check again.

apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -303,15 +314,13 @@ static Options buildOptions() {
...
 
         OptionGroup addLicenseGroup = new OptionGroup();
-        String addLicenseDesc = "Add the default license header to any file with an unknown license that is not in the exclusion list. "
-                + "By default new files will be created with the license header, "
-                + "to force the modification of existing files use the --force option.";
-
         // RAT-85/RAT-203: Deprecated! added only for convenience and for backwards
         // compatibility
-        Option addLicence = new Option(ADD_OLD, "addLicence", false, addLicenseDesc);
+        Option addLicence = new Option(ADD_OLD, false, "(deprecated) Add the default license header to any file with an unknown license.  Use '-A' or ---addLicense instead.");
```
ottlinger, line 319: Should we rename the internal variable to ADD_DEPRECATED (instead of OLD)?

apache-rat/src/site/apt/index.apt.vm
ottlinger: Now it is visible for me as well. Thanks for verifying!

src/site/apt/index.apt.vm
ottlinger: fyi: I'll will try to add a link on the the branch ....

ottlinger: Will commit directly to your branch - hope this is okay,  @Claudenw 
fyi:
https://maven.apache.org/doxia/references/apt-format.html

apache-rat-core/src/main/java/org/apache/rat/Report.java
Claudenw: I was thinking about this this AM.  We can change it to ADD_DEPRECATED,  I see that as a minor change.

But since we are exposing the stylesheets by name,  I would like to rename `plain-rat` to `rat`  and create one called `xml` that will produce a pretty printed XML output.  We can then remove the `-x` option and consider if we want to support running multiple transforms against the output in a single run using multiple `-s` options.

What do you think of the multiple `-s` option idea and do you agree with the renaming of `plain-rat` to `rat`?

ottlinger:
@Claudenw if it's fine for you feel free to merge this PR :)

PR #228, opened 2024-03-22
RAT-366: switch to single match call

Claudenw:
Simplification of match processing.  Minor performance improvements. Code path improvement and simplification.

This change removes the State tracking for matches and switches from a line-by-line approach to a text-block approach.  The change is that we read all the lines that we are going to process before we do any tests and then we call the matchers.

This means that this change will hit every matcher in the system as well as the match handling code.  However, the code path is significantly simplified as a result of removing the State tracking necessary for the line-by-line approach.

As you can see most of the code changes are the removal of state and some reduction in ancillary tracking of the text seen in some matchers.


Claudenw:
This was code I developed while looking at the performance problem a while back.  I did not include these changes then as there was a much smaller change necessary to effect the desired improvement.  This is a refactor/code cleanup exercise.

PR #230, opened 2024-03-23
RAT-367: Enable running old junit tests during the build with junit5

nielsbasjes:
This fixes https://issues.apache.org/jira/browse/RAT-367 and now also the older tests are run on each build.


ottlinger:
Thanks for the catch, I'll add a changelog entry accordingly.

PR #233, opened 2024-04-01
RAT-355 and RAT-366 fixes in one package

Claudenw:
Contains all the changes for RAT-355 and RAT-366 as one package.

The two issues were intertwined to the point where multiple files would be updated  by both changes and in conflicting ways.  This branch contains the changes from both.

The main changes are:

- MetaData now tracks all the licenses that are applicable to the document and allows us to print out the discovered licenses and whether or not they are approved.
- Matches checked in a single call, not line by line.  This should noticeably  speed up processing for some cases.  This also removed all the "state" code that was inserted when the "not" matcher was added.
- The Configuration file can be dumped as part of the XML report.  There are no styles that report on this information yet, but it is available in the XML output.
- To assist in parsing and building matcher a `ConfigComponent` annotation was added, as was a `Component` interface that is now the root for `IHeaderMatcher` and `ILicense`, a `Description` class that describe the Components, and a `DescriptionBuilder` that builds the Descriptions.  This results in much simpler `IHeaderMatcher` and `IHeaderMatcher.Builder` code. 

I am currently working on cleaning up the javadoc but I do not expect any further executable code changes.


apache-rat-core/src/main/java/org/apache/rat/config/parameters/Component.java
```diff
@@ -0,0 +1,48 @@
...
+    /** Types of components */
+    public enum Type {
+        /** A License, the top level component. May not be used as a child of any component type. */
+        License,
+        /** A Matcher */
+        Matcher,
+        /** A Parameter for example the "id" parameter found in every component */
+        Parameter,
+        /** A parameter without a name.  Generally this is a value for the item.  For example the enclosed matchers in the "or" license. */
+        Unlabeled,
```
Claudenw, line 35: this type mixes presentation with functionality.  It should be removed and specific display issues addressed in the format serializers/deserializers

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
```diff
@@ -0,0 +1,285 @@
+/*
```
Claudenw, line 1: The functionality in this class should be in an XMLConfigurationWriter that this class will use to deserialize the configuration.

apache-rat-core/src/main/java/org/apache/rat/Documentation.java
```diff
@@ -0,0 +1,152 @@
+/*
```
Claudenw, line 1: This class should be placed in the apache-rat-tools project and documented .

Claudenw (commented):
The Ant tasks and Maven builders should be generated from the configuration information.  This will ensure that 3rd party matchers can be properly integrated into the various build systems and will provide a pattern for 3rd parties to add RAT functionality to other build systems.

apache-rat-core/src/main/java/org/apache/rat/config/parameters/Component.java
```diff
@@ -0,0 +1,48 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Interface that marks a configuration component like a License or Matcher.
+ *
+ */
+public interface Component {
+    /** Types of components */
```
Claudenw, line 26: Remove this interface and move the `getDescription` method to IHeaderMatcher interface.
Move the enclosed Type enum into Description class.

apache-rat-core/src/main/java/org/apache/rat/api/RatException.java
```diff
@@ -19,9 +19,20 @@
...
 
 public class RatException extends Exception {
+    
+    public static RatException asRatException(Exception e) {
+        return e instanceof RatException ? (RatException) e : new RatException(e);
+    }
 
     private static final long serialVersionUID = 4940711222435919034L;
 
+    public static RatException makeInstance(Exception e) {
```
Claudenw, line 29: remove this infavor of the `asRatException` method

Claudenw:
**Builder changes:**
 * Matcher Builders now return specific type, not the IHeaderMatcher but a subclass of that.
 
**Matcher changes:**
 * removed state tracking code.
 * added ConfigComponent annotations to describe the matcher and it properties.
 * in some cases added getter methods to retrieve property values.

**Metadata / ClaimStatistic changes**
  Complete rewrite to simplify setting and getting values (`Metadata`) and to track those values (`ClaimStatistic`)
  
**AbstractClaimReporter**
 * Changed to use Document.Type and Metadata.
 * Significant simplification.
 
**ClaimAggregator**
 * modified to use Document.Type and new ClaimStatistic api.
 
**SimpleXmlClaimReporter **

 * modified to handle multiple licenses for single document.
 
 
**ILicenseFamily  changes:**

Added definitions for Generated and Unknown family categories so that they are defined in one place.

**ILicenseChanges**

Removed `derivedfrom` attribute as it was added recently and unused.
  
**IXmlWriter / XMLWriter changes**

 * added cdata method
 
**Should change**

Remove `ToNameTransformer` as it is not used and not part of the API

apache-rat-core/src/main/java/org/apache/rat/Documentation.java
```diff
@@ -0,0 +1,152 @@
...
+            case BuilderParam:
+                writer.write(String.format("'%s' %s (Datatype: %s%s)%n", d.getCommonName(), d.getDescription(),
+                        d.isCollection() ? "Collection of " : "", d.getChildType().getSimpleName()));
+                break;
+            case Unlabeled:
+                writer.write(String.format("'%s name' %s (Datatype: %s%s)%n", d.getCommonName(), d.getDescription(),
+                        d.isCollection() ? "Collection of " : "", d.getChildType().getSimpleName()));
+                break;
+            case License:
+                // do noting
```
ottlinger, line 115: Typo ;) nothing - i though of moving it to a default block .... when reading it first.

apache-rat-core/src/main/java/org/apache/rat/Documentation.java
```diff
@@ -0,0 +1,152 @@
...
+                writer.write(String.format("%s %s %n", d.isCollection() ? "Collection of " : "A ",
+                        d.getChildType().getSimpleName()));
+                printChildren(indent + 2, d.getChildren(), writer);
+                break;
+            }
+        }
+    }
+
+    /**
+     * Creates the documentation.  Writes to the outpus specified by the -o or --out option.  Defaults to System.out.
```
ottlinger, line 128: Typo: output

apache-rat-core/src/main/java/org/apache/rat/analysis/DocumentHeaderAnalyser.java
```diff
@@ -18,50 +18,50 @@
...
-    public void analyse(Document document) throws RatDocumentAnalysisException {
+    public void analyse(Document document) {
         try (Reader reader = document.reader()) {
             log.debug(format("Processing: %s", document));
-            HeaderCheckWorker worker = new HeaderCheckWorker(reader, license, document);
+            HeaderCheckWorker worker = new HeaderCheckWorker(reader, licenses, document);
             worker.read();
         } catch (IOException e) {
-            throw new RatDocumentAnalysisException("Cannot read header", e);
+            log.warn(String.format("Can not read header of %s", document));
```
ottlinger, line 60: cannot?

apache-rat-core/src/main/java/org/apache/rat/analysis/DocumentHeaderAnalyser.java
```diff
@@ -18,50 +18,50 @@
...
+            document.getMetaData().setDocumentType(Document.Type.unknown);
         } catch (RatHeaderAnalysisException e) {
-            throw new RatDocumentAnalysisException("Cannot analyse header", e);
+            log.warn(String.format("Can not process header of %s", document));
```
ottlinger, line 63: cannot?

apache-rat-core/src/main/java/org/apache/rat/analysis/HeaderCheckWorker.java
```diff
@@ -47,98 +53,103 @@ class HeaderCheckWorker {
...
+    public static IHeaders readHeader(BufferedReader reader, int numberOfLines) throws IOException {
+        final StringBuilder headers = new StringBuilder();
+        int headerLinesRead = 0;
+        String line;
+
+        while (headerLinesRead < numberOfLines && (line = reader.readLine()) != null) {
+            headers.append(line).append("\n");
+        }
+        final String raw = headers.toString();
+        final String pruned = FullTextMatcher.prune(raw).toLowerCase(Locale.ENGLISH);
```
ottlinger, line 78: will this bring in problems with non-english stuff? I remember there is a bug in the bkclog that a chinese document is detected as binary .... just wondering

apache-rat-core/src/main/java/org/apache/rat/analysis/HeaderCheckWorker.java
```diff
@@ -47,98 +53,103 @@ class HeaderCheckWorker {
...
+            }
+
+            @Override
+            public String pruned() {
+                return pruned;
+            }
+
+            @Override
+            public String toString() {
+                return "HeaderCheckWorker";
```
ottlinger, line 92: Should we use HeaderCheckWorker.class.getSimpleName() to make it refactoring-safer?

apache-rat-core/src/main/java/org/apache/rat/analysis/IHeaderMatcher.java
```diff
@@ -110,10 +71,39 @@ public String toString() {
...
 
+        /**
+         * Gets the class that is build by this builder.
+         * @return The class that is build by this builder.
+         */
+        default Class<?> builtClass() throws SecurityException {
+            try {
+                return this.getClass().getMethod("build").getReturnType();
+            } catch (NoSuchMethodException | SecurityException e) {
+                throw new IllegalStateException("the 'build' method of the Builder interface must alwasy be public");
```
ottlinger, line 89: typo: always

apache-rat-core/src/main/java/org/apache/rat/analysis/IHeaderMatcher.java
```diff
@@ -110,10 +71,39 @@ public String toString() {
...
+
+        /**
+         * Gets the Description for this builder.
+         * @return The description of the builder 
+         */
+        default Description getDescription() {
+            Class<?> clazz = builtClass();
+            if (clazz == IHeaderMatcher.class) {
+                throw new ImplementationException(String.format(
+                        "Class %s must implement built() method to return a child class of IHeaderMatcher",
```
ottlinger, line 101: typo: build()?

apache-rat-core/src/main/java/org/apache/rat/analysis/IHeaders.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface IHeaders {
+    /**
+     * Gets raw header as found in the document. 
+     * @return the raw header as read from the file.
+     */
+    public String raw();
+
+    /**
+     * Gets the pruned header.  This is the header with all non letter and non number characters removed and the string converted to lower case..
```
ottlinger, line 32: typo: ..

apache-rat-core/src/main/java/org/apache/rat/analysis/LicenseCollection.java
```diff
@@ -22,33 +22,37 @@
...
 
 /**
- * A collection of ILicenses that acts as a single License for purposes of Analysis.
+ * A collection of ILicenses that acts as a single License for purposes of
+ * Analysis.
  * <p>
- * This class process each license in turn on each {@code matches(String)} call.  When a match is found the 
- * ILicenseFamily for the matching license is captured and used as the family for this license. If no matching 
- * license has been found the default {@code dummy} license category is used.
+ * This class process each license in turn on each {@code matches(String)} call.
```
ottlinger, line 33: typo: processes instead of process

apache-rat-core/src/main/java/org/apache/rat/analysis/LicenseCollection.java
```diff
@@ -22,33 +22,37 @@
...
+ * When a match is found the ILicenseFamily for the matching license is captured
+ * and used as the family for this license. If no matching license has been
+ * found the default {@code dummy} license category is used.
  */
+
 class LicenseCollection extends AbstractMatcherContainer implements ILicense {
 
     private static final ILicenseFamily DEFAULT = ILicenseFamily.builder().setLicenseFamilyCategory("Dummy")
```
ottlinger, line 41: Not sure how this dummy is handled in the report, would NONE be clearer here as the family does actually not do anything?

apache-rat-core/src/main/java/org/apache/rat/analysis/matchers/CopyrightMatcher.java
```diff
@@ -57,32 +63,61 @@ public class CopyrightMatcher extends AbstractSimpleMatcher {
 
     private final Pattern dateOwnerPattern;
     private final Pattern ownerDatePattern;
+    @ConfigComponent(type = Component.Type.Parameter, desc = "The initial date of the copyright")
+    private final String start;
+    @ConfigComponent(type = Component.Type.Parameter, desc = "The last date the copyright we modifed")
```
ottlinger, line 68: ypo: was instead of we

apache-rat-core/src/main/java/org/apache/rat/analysis/matchers/CopyrightMatcher.java
```diff
@@ -104,13 +139,37 @@ public CopyrightMatcher(String id, String start, String stop, String owner) {
...
+    }
+
     @Override
-    protected boolean doMatch(String line) {
-        String lowerLine = line.toLowerCase();
-        if (lowerLine.contains("copyright") || lowerLine.contains("(c)") || line.contains("©")) {
-            Matcher matcher = COPYRIGHT_PATTERN.matcher(line);
+    public boolean matches(IHeaders headers) {
+        String lowerLine = headers.raw().toLowerCase();
+        if (lowerLine.contains("copyright") || lowerLine.contains("(c)") || lowerLine.contains("©")) {
```
ottlinger, line 169: Would it make sense to match &copy; as well in case of HTML files?

apache-rat-core/src/main/java/org/apache/rat/analysis/matchers/OrMatcher.java
```diff
@@ -21,75 +21,44 @@
...
+import org.apache.rat.analysis.IHeaders;
+import org.apache.rat.config.parameters.Component;
+import org.apache.rat.config.parameters.ConfigComponent;
 
 /**
- * A matcher that performs a logical {@code OR} across all the contained matchers.
+ * A matcher that performs a logical {@code OR} across all the contained
+ * matchers.
  */
+@ConfigComponent(type = Component.Type.Matcher, name = "any", desc = "Returns true if at least one of the enclosed matchers return true.")
```
ottlinger, line 32: "returns true"?

apache-rat-core/src/main/java/org/apache/rat/analysis/matchers/SimpleTextMatcher.java
```diff
@@ -19,36 +19,49 @@
...
+import org.apache.rat.analysis.IHeaders;
+import org.apache.rat.config.parameters.Component;
+import org.apache.rat.config.parameters.ConfigComponent;
 
 /**
  * A simple text matching IHeaderMatcher implementation.
  */
-public class SimpleTextMatcher extends AbstractSimpleMatcher {
-    private final String pattern;
+@ConfigComponent(type = Component.Type.Matcher, name = "text", desc = "Matches the enclosed the text")
```
ottlinger, line 29: Typo "the enclosed text"

apache-rat-core/src/main/java/org/apache/rat/api/Document.java
```diff
@@ -24,7 +24,26 @@
...
+ * The representation of a document being scanned.
+ */
 public interface Document {
+    /**
+     * An enumeraton of document types.
+     */
+    enum Type {
+        /** A generated document. */
+        generated, 
+        /** An unknowns document type. */
```
ottlinger, line 37: Typo: An unknown?!

apache-rat-core/src/main/java/org/apache/rat/api/Document.java
```diff
@@ -24,7 +24,26 @@
...
+/**
+ * The representation of a document being scanned.
+ */
 public interface Document {
+    /**
+     * An enumeraton of document types.
+     */
+    enum Type {
+        /** A generated document. */
+        generated,
```
ottlinger, line 36: Shouldn't we be UPPERCASE here to comply with Java standards?

apache-rat-core/src/main/java/org/apache/rat/api/MetaData.java
```diff
@@ -18,340 +18,128 @@
...
-        this.data = new ArrayList<>(16);
+        this.matchedLicenses = new TreeSet<>();
+        this.approvedLicenses = new HashSet<>();
     }
-    
+
     /**
-     * Add the license information to the metadata.
-     * @param license the license to add metadata for.
+     * Determines if a mathcing license has been detected.
```
ottlinger, line 52: Typo: matching

apache-rat-core/src/main/java/org/apache/rat/api/MetaData.java
```diff
@@ -18,340 +18,128 @@
...
     }
-    
+
     /**
-     * Puts in a new datum replacing any existing data.
-     * Any current data matching the name are removed.
-     * @param datum not null
-     * @see #add(org.apache.rat.api.MetaData.Datum)
+     * Gets the stream of approved licenses that have been matched.
+     * @return the stream of approved licenses theat have been matched.
```
ottlinger, line 78: Typo: that

apache-rat-core/src/main/java/org/apache/rat/config/parameters/Component.java
```diff
@@ -0,0 +1,48 @@
...
+    public enum Type {
```
ottlinger, line 27: Uppercasing?

apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
```diff
@@ -0,0 +1,321 @@
...
+     */
+    private final String name;
+    /** The description for the component */
+    private final String desc;
+    /** The class of the getter/setter parameter */
+    private final Class<?> childClass;
+    /** True if the getter/setter expects a collection of childClass objects */
+    private final boolean isCollection;
+    /**
+     * a map of name to Description for all the components that are children the
```
ottlinger, line 52: Typo: children of the ....

apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
```diff
@@ -0,0 +1,321 @@
...
+    /**
+     * Constructor.
+     * 
+     * @param type the type of the component.
+     * @param name the name of the component.
+     * @param desc the description of the component.
+     * @param isCollection true if the getter/setter expects a collection
+     * @param childClass the class for expected for the getter/setter.
+     * @param children the collection of descriptions for all the components that
+     * are children the described component.
```
ottlinger, line 66: Typo: of the described component

apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
```diff
@@ -0,0 +1,321 @@
...
+     * Get the isCollection flag.
+     * 
+     * @return true if this is a collection.
+     */
+    public boolean isCollection() {
+        return isCollection;
+    }
+
+    /**
+     * Get the class of the objcts for the getter/setter.
```
ottlinger, line 117: Typo: objects

apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
```diff
@@ -0,0 +1,321 @@
...
+     * The common name for the component. Set by ConfigComponent.name() or
+     * class/field name.
+     */
+    private final String name;
+    /** The description for the component */
+    private final String desc;
+    /** The class of the getter/setter parameter */
+    private final Class<?> childClass;
+    /** True if the getter/setter expects a collection of childClass objects */
+    private final boolean isCollection;
```
ottlinger, line 50: When I read this the first time I instantly thought of isRecursive(), but it would imply a great many changes.

apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    /**
+     * Constructor
+     * 
+     * @param configComponent the configuration component
+     * @param isCollection the collection flag.
+     * @param childClass the type of object that the method getter/setter expects.
+     * @param children the collection of descriptions for all the components that
+     * are children the described component.
```
ottlinger, line 90: Typo: of the described component

apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
```diff
@@ -0,0 +1,321 @@
...
+    private void callSetter(Log log, Description description, IHeaderMatcher.Builder builder, String value) {
+        try {
+            description.setter(builder.getClass()).invoke(builder, value);
+        } catch (NoSuchMethodException e) {
+            String msg = String.format("No setter for '%s' on %s", description.getCommonName(), builder.getClass().getCanonicalName()); 
+            log.error(msg);
+            throw new ConfigurationException(msg);
+        } catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException
+                |  SecurityException e) {
+            String msg = String.format("can not call setter for '%s' on %s", description.getCommonName(), builder.getClass().getCanonicalName());
```
ottlinger, line 249: Unable to instead of cannot? Not sure where this message is shown.

apache-rat-core/src/main/java/org/apache/rat/analysis/LicenseCollection.java
Claudenw: Class should have been deleted

apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
Claudenw: On a 3rd party class if the setter is not public this exception is thrown, or if there is a security manager that restricts the call.  Changed the text to "Unable to call..."

apache-rat-core/src/main/java/org/apache/rat/configuration/XMLConfigurationReader.java
```diff
@@ -279,14 +334,13 @@ private ILicense parseLicense(Node licenseNode) {
         StringBuilder notesBuilder = new StringBuilder();
         nodeListConsumer(licenseNode.getChildNodes(), x -> {
             if (x.getNodeType() == Node.ELEMENT_NODE) {
-                if (x.getNodeName().equals(NOTE)) {
+                if (NOTE.equals(x.getNodeName())) {
                     notesBuilder.append(x.getTextContent()).append("\n");
```
ottlinger: Would System.lineSeparator() be more correct here?

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/ChildContainerBuilder.java
```diff
@@ -52,39 +52,42 @@ protected ChildContainerBuilder() {
...
-                BufferedReader buffer = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8));
-                String txt;
-                while (null != (txt = buffer.readLine())) {
-                    txt = txt.trim();
-                    if (StringUtils.isNotBlank(txt)) {
-                        children.add(Builder.text().setText(txt));
-                    }
+        URL url = this.getClass().getResource(resourceName);
+        try (final InputStream in = url.openStream()) {
+            BufferedReader buffer = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8));
```
ottlinger, line 63: Would we want to put the buffered reader into the tryWithResources as well?

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/ChildContainerBuilder.java
```diff
@@ -99,11 +102,11 @@ public AbstractBuilder add(Collection<IHeaderMatcher.Builder> children) {
...
         return children.stream().map(IHeaderMatcher.Builder::build).collect(Collectors.toList());
     }
-    
+
     @Override
     public String toString() {
-        StringBuilder sb = new StringBuilder(this.getClass().getSimpleName()).append( ":");
-        children.stream().map(Object::toString).forEach( x -> sb.append("\n").append(x));
+        StringBuilder sb = new StringBuilder(this.getClass().getSimpleName()).append(":");
+        children.stream().map(Object::toString).forEach(x -> sb.append("\n").append(x));
```
ottlinger, line 113: Not sure if it is a problem on Windows to have Linux file endings ... otherwise System.lineSeparator() again.

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/MatcherRefBuilder.java
```diff
@@ -20,19 +20,26 @@
...
 /**
- * A reference matching Matcher builder.
- * <p>
- * This class stores a matcher id as a reference to the matcher.  It also has a map of matcher ids to the matcher
- * instances.  When build is called the matcher reference is looked up in the map.  If it is found then it is returned
- * value from the {@code build()} call.  If the reference is not located then a IHeaderMatcherProxy is returned.
- * the IHeaderMatcherProxy is resolved in a later configuration construction phase.
+ * A reference matching Matcher builder. <p> This class stores a matcher id as a
+ * reference to the matcher. It also has a map of matcher ids to the matcher
+ * instances. When build is called the matcher reference is looked up in the
```
ottlinger: When {@code build()} is called ....

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/MatcherRefBuilder.java
```diff
@@ -20,19 +20,26 @@
...
+ * map. If it is found then it is returned value from the {@code build()} call.
```
ottlinger: If it is found, its value is returned from the {@code build()} call.

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/MatcherRefBuilder.java
```diff
@@ -56,31 +68,53 @@ public MatcherRefBuilder setMatchers(Map<String, IHeaderMatcher> matchers) {
...
     }
-    
+
     /**
-     * A class that is a proxy to the actual matcher.  It retrieves the actual matcher from the map of
-     * matcher ids to matcher instances one the first use of the matcher.  This allows earlier read matchers
-     * to reference later constructed matchers as long as all the matchers are constructed before the earlier one is 
-     * used.
+     * A class that is a proxy to the actual matcher. It retrieves the actual
+     * matcher from the map of matcher ids to matcher instances one the first use of
```
ottlinger, line 89: on the first use of .....

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/NotBuilder.java
```diff
@@ -19,24 +19,23 @@
...
             throw new ConfigurationException("'not' type matcher requires one and only one enclosed matcher");
         }
         return new NotMatcher(getId(), children.get(0).build());
     }
-    
+
     @Override
     public String toString() {
-        return String.format( "NotBuilder: %s", !children.isEmpty() ? children.get(0) : null );
+        return String.format("NotBuilder: %s", !children.isEmpty() ? children.get(0) : null);
```
ottlinger, line 40: What is the semantics of an empty NotBuilder? Is it "not empty"? 
Personally I'd expect an empty notBuilder to be in an illegal state

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/SpdxBuilder.java
```diff
@@ -32,22 +31,22 @@ public class SpdxBuilder extends AbstractBuilder {
...
     /**
      * sets the name for the SPDX matcher
-     * @param name
-     * @return
+     * @param name The text that follows the colon ':' in the SPDX tag.
+     * @return this builder for chaining.
      */
     public SpdxBuilder setName(String name) {
-        Objects.requireNonNull(name, "name must not be null");
+        Objects.requireNonNull(name, "spdx name must not be null");
```
ottlinger, line 41: Does SPDX require uppercasing, if so we could enforce it here and in the constructor?

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/TextBuilder.java
```diff
@@ -22,27 +22,24 @@
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.rat.ConfigurationException;
-import org.apache.rat.analysis.IHeaderMatcher;
 import org.apache.rat.analysis.matchers.FullTextMatcher;
 import org.apache.rat.analysis.matchers.SimpleTextMatcher;
 
 /**
  * Builds text based matcher based on the complexity of the text to match.
```
ottlinger, line 27: Builds a text based matcher?!

apache-rat-core/src/main/java/org/apache/rat/document/impl/ArchiveEntryDocument.java
```diff
@@ -37,44 +36,44 @@ public class ArchiveEntryDocument implements Document {
...
     }
 
+    @Override
     public InputStream inputStream() throws IOException {
         return new ByteArrayInputStream(contents);
     }
 
+    @Override
     public boolean isComposite() {
         return DocumentImplUtils.isZipStream(new ByteArrayInputStream(contents));
```
ottlinger, line 62: Could we change the interface to internally wrap/add the ByteArrayInputStream in isZipStream() and autoclose it with a tryWithresources construct?

apache-rat-core/src/main/java/org/apache/rat/license/ILicense.java
```diff
@@ -22,42 +22,58 @@
...
      * @return An ILicense.Builder instance.
      */
     static ILicense.Builder builder() {
         return new Builder();
     }
 
     /**
-     * @return The comparator for used to sort Licenses. 
+     * Gets the comparator for comparing licenses.
+     * @return The comparator for used to sort Licenses.
```
ottlinger, line 80: The comparator that is used for sorting licenses.

apache-rat-core/src/main/java/org/apache/rat/license/LicenseSetFactory.java
```diff
@@ -220,5 +221,5 @@ public static ILicense search(ILicense target, SortedSet<ILicense> licenses) {
         SortedSet<ILicense> part = licenses.tailSet(target);
         return (!part.isEmpty() && part.first().compareTo(target) == 0) ? part.first() : null;
```
ottlinger, line 222: Can target be null? Should we use part().first().equals(target) instead of the compareTo==0?

apache-rat-core/src/main/java/org/apache/rat/policy/DefaultPolicy.java
```diff
@@ -45,6 +46,7 @@ public DefaultPolicy(final Collection<ILicenseFamily> approvedLicenseFamilies) {
 
     /**
      * adds an ILicenseFamily to the list of approved licenses.
```
ottlinger, line 48: Typo: Adds an ....

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    private void writeAttribute(Description description, Component component) throws IOException {
+        String paramValue = description.getParamValue(configuration.getLog(), component);
+        if (paramValue != null) {
+            writer.attribute(description.getCommonName(), paramValue);
+        }
+    }
+
+    /* package privete for testing */
```
ottlinger: "@VisibleForTesting" ? Not sure if we already have this annotation. I know it from Guava.

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
```diff
@@ -0,0 +1,285 @@
...
+                    // if we have seen the ID before just put a reference to the other one.
+                    if (matchers.contains(idValue.toString())) {
+                        component = new MatcherRefBuilder.IHeaderMatcherProxy(idValue.toString(), null);
+                        description = component.getDescription();
+                    } else {
+                        matchers.add(idValue.toString());
+                    }
+                }
+                Optional<Description> resource = description.childrenOfType(Type.Parameter).stream()
+                        .filter(i -> "resource".equals(i.getCommonName())).findFirst();
```
ottlinger: Should we extract this into XMLConfigurationReader.RESOURCE here?

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
```diff
@@ -0,0 +1,285 @@
...
+        if (paramValue != null) {
+            writer.attribute(description.getCommonName(), paramValue);
+        }
+    }
+
+    /* package privete for testing */
+    void writeDescription(Description description, Component component) throws RatException {
+        try {
+            switch (description.getType()) {
+            case Matcher:
```
ottlinger: Reading the usage of these enums I'd prefer to use Uppercase values to not mix it with class names (or confuse readers of the sources ;))

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
```diff
@@ -0,0 +1,285 @@
...
+                writeComment(description);
+                writer.openElement(description.getCommonName());
+                writeChildren(description, component);
+                writer.closeElement();
+                break;
+            case License:
+                writer.openElement(XMLConfigurationReader.LICENSE);
+                Description notes = null;
+                for (Description desc : description.childrenOfType(Component.Type.Parameter)) {
+                    if ("notes".equals(desc.getCommonName())) {
```
ottlinger: Extract as constant as other occurences above?

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
```diff
@@ -0,0 +1,285 @@
...
+                        notes = desc;
+                    } else {
+                        writeAttribute(desc, component);
+                    }
+                }
+                for (Description desc : description.childrenOfType(Component.Type.BuilderParam)) {
+                    if ("family".equals(desc.getCommonName())) {
```
ottlinger: Extract as constant as other occurences above?

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
```diff
@@ -0,0 +1,285 @@
...
+                    writeContent(notes, component);
+                    writer.closeElement();
+                }
+                for (Description desc : description.childrenOfType(Component.Type.Unlabeled)) {
+                    writeDescription(desc, component);
+                }
+                writer.closeElement();
+                break;
+            case Parameter:
+                if ("id".equals(description.getCommonName())) {
```
ottlinger: Extract as constant as other occurences above?

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
```diff
@@ -0,0 +1,285 @@
...
+                    try {
+                        String idValue = description.getParamValue(configuration.getLog(), component);
+                        // if a UUID skip it.
+                        if (idValue != null) {
+                            UUID.fromString(idValue.toString());
+                            return;
```
ottlinger: Does this leave broken XML? Maybe I do not understand it correctly ..... or does it just not write any thing as parameters do not have descriptions?

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
```diff
@@ -0,0 +1,285 @@
...
+                    }
+                }
+                writeAttribute(description, component);
+                break;
+            case Unlabeled:
+                try {
+                    Object obj = description.getter(component.getClass()).invoke(component);
+                    if (obj instanceof Iterable) {
+                        for (Object o2 : (Iterable<?>) obj) {
+                            processUnlabled(o2);
```
ottlinger: Typo in method name - processUnlabeled() instead?

apache-rat-core/src/main/java/org/apache/rat/report/claim/impl/ClaimAggregator.java
```diff
@@ -19,84 +19,80 @@
...
-    protected void handleApprovedLicenseClaim(String licenseApproved) {
-        if (MetaData.RAT_APPROVED_LICENSE_VALUE_TRUE.equals(licenseApproved)) {
-            numApproved++;
+    protected void handleApprovedLicenseClaim(MetaData metadata) {
+        incValueMap(statistic.getCounterMap(), ClaimStatistic.Counter.Approved, metadata.approvedLicenses().count());
+        incValueMap(statistic.getCounterMap(), ClaimStatistic.Counter.Unapproved,
+                metadata.unapprovedLicenses().count());
+    }
+
+    private void incValueMap(Map<Counter, int[]> map, Counter key, long value) {
```
ottlinger, line 68: Why do we use long here and shorten it to integer internally?

apache-rat-core/src/main/java/org/apache/rat/report/claim/impl/xml/SimpleXmlClaimReporter.java
```diff
@@ -140,9 +100,9 @@ public void startReport() throws RatException {
     @Override
     public void endReport() throws RatException {
         try {
-            writer.closeDocument();
+            writer.closeElement();
         } catch (IOException e) {
-            throw new RatException("Cannot close last element", e);
+            throw new RatException("Cannot open start element", e);
```
ottlinger, line 105: "Cannot close start element"? endReport() should not open the start element here.

apache-rat-core/src/main/java/org/apache/rat/report/xml/writer/IXmlWriter.java
```diff
@@ -66,8 +76,8 @@ public interface IXmlWriter extends AutoCloseable {
     
     /**
      * Writes content.
-     * Calling this method will automatically 
-     * Note that this method does not use CDATA.
+     * Note that this method does not support CDATA.
+     * This method automatically excapes characters.
```
ottlinger, line 80: Typo: escapes characters

apache-rat-core/src/main/java/org/apache/rat/report/xml/writer/impl/base/XmlWriter.java
```diff
@@ -514,6 +526,17 @@ private void writeAttributeContent(CharSequence content) throws IOException {
...
+                throw new OperationNotAllowedException("Root element has already been closed.");
+            } 
+            throw new OperationNotAllowedException("An element must be opened before content can be written.");
+        }
+        if (inElement) {
+            writer.write('>');
+        }
+    }
     /**
      * Writes content. Calling this method will automatically Note that this method
```
ottlinger, line 541: Calling this method will automatically .... seems like a copy&paste thing - 
Do you want to add info that the element is closed automtically?

apache-rat-core/src/main/java/org/apache/rat/report/xml/writer/impl/base/XmlWriter.java
```diff
@@ -525,24 +548,36 @@ private void writeAttributeContent(CharSequence content) throws IOException {
...
+        writeEscaped(content, false);
         inElement = false;
         return this;
     }
 
-    private void writeBodyContent(final CharSequence content) throws IOException {
-        writeEscaped(content, false);
+    
+    /**
+     * Writes content. Calling this method will automatically Note that this method
```
ottlinger, line 559: Calling this method will automatically .... seems like a copy&paste thing - 
Do you want to add info that the element is closed automtically?

Pls add a linebreak before "Note that ..."

apache-rat-core/src/main/java/org/apache/rat/report/xml/writer/impl/base/XmlWriter.java
```diff
@@ -525,24 +548,36 @@ private void writeAttributeContent(CharSequence content) throws IOException {
...
+     * does not use CDATA.
+     *
+     * @param content the content to write
+     * @return this object
+     * @throws OperationNotAllowedException if called before any call to
+     * {@link #openElement} or after the first element has been closed
+     */
+    @Override
+    public IXmlWriter cdata(CharSequence content) throws IOException {
```
ottlinger, line 568: Pls reformat this method - thx.

apache-rat-core/src/main/java/org/apache/rat/report/xml/writer/impl/base/XmlWriter.java
```diff
@@ -525,24 +548,36 @@ private void writeAttributeContent(CharSequence content) throws IOException {
...
+        prepareForData();
+        StringBuilder sb = new StringBuilder(content);
+        int found;
+        while (-1 != (found = sb.indexOf("]]>"))) {
+            sb.replace(found, found+3, "<!- Rat removed CDATA closure here ->");
```
ottlinger, line 573: Should this be a real comment <!-- or is this a special marker?

apache-rat-core/src/main/java/org/apache/rat/utils/Log.java
```diff
@@ -59,51 +59,105 @@ public enum Level {
...
+     * @param message the message for the log
+     * @param throwable the throwable
+     */
+    default void log(Level level, String message, Throwable throwable) {       
         StringWriter writer = new StringWriter(500);
         PrintWriter pWriter = new PrintWriter(writer);
         pWriter.print(message);
         pWriter.print(System.lineSeparator());
-        t.printStackTrace(pWriter);
+        throwable.printStackTrace(pWriter);
```
ottlinger, line 114: The docs says that printStackTrace() prints to StandardErr. I thought we want to log this as part of the given message or is it okay to write a logLevel in DEBUG that outputs the stacktrace on ERROR?

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
```diff
@@ -53,42 +53,58 @@ Files with unapproved licenses:
...
 *****************************************************
-  Files with Apache License headers will be marked AL
-  Binary files (which do not require any license headers) will be marked B
-  Compressed archives will be marked A
-  Notices, licenses etc. will be marked N
+  Documents with unapproved licenses will start with a '!'
+  The next character identifies the document type.
+   
+   char         type
+    a       Archive file
```
ottlinger: Should we use uppercase chars to identify the document's type to be more compliant with current RAT reports?

jbonofre:
Sorry guys, I'm late on this one. I will take a look today.

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/ChildContainerBuilder.java
Claudenw: All output \n have been converted to System.lineSeparator() or String.format based "%n"

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/NotBuilder.java
Claudenw: The builder can be in a state where the enclosed matcher has not yet been constructed.  However, when the `build()` method is called it will throw an exception if the enclosed matcher is not defined.  

I suppose the NotBuilder is in a Schrodengers Cat state if the matcher has not been defined.

apache-rat-core/src/main/java/org/apache/rat/document/impl/ArchiveEntryDocument.java
Claudenw: Actually, this file has no technical changes and should be reverted.  But in answer to your question yes, and it is an area I will be looking into shortly.  I want us to use the SPDX license matching code to see if we can match LICENSE files and I want to do that within archive files.

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/TextBuilder.java
Claudenw: reworked.

apache-rat-core/src/main/java/org/apache/rat/license/LicenseSetFactory.java
Claudenw: ILicense does not override `equals` but does implement comparable, so `equals` may not always work as expected.  A null target will throw a null pointer exception.

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
Claudenw: We do not.  I didn't want to bring it in unless we had a good reason to.  Do we have tooling that would use it?

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
Claudenw: Extracted to XMLConfig

apache-rat-core/src/main/java/org/apache/rat/configuration/builders/SpdxBuilder.java
Claudenw: We are only looking for the identifier for the license [1] the search for the SPDX part is automatic.  But we should be consistent in the documentation.  Change to 'SPDX' made in all found documentation.

[1] https://spdx.org/licenses/

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
Claudenw: All enums have been changed to upper case.

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
Claudenw: all constants have been extracted.

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
Claudenw: All IHeaderMatcher instances have to have a unique ID.  If none is specified we create a time based UUID for the processing run.  This verifies that a UUID is not output as the ID into the XML document.  I suppose there should be a warning to users not to use UUIDs ad matcher ids.  The current code, in XMLConfigurationWriter, handles this a bit differently but it still removes the ID from the XML on output if it is a UUID.

apache-rat-core/src/main/java/org/apache/rat/report/ConfigurationReport.java
Claudenw: unlabelled has been removed as a type.

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
```diff
@@ -19,100 +19,62 @@
...
-    }
-
-    /**
-     * @return Returns the number of generated files.
-     */
-    public int getNumGenerated() {
-        return numGenerated;
-    }
+    /** The counter types */
+    public enum Counter {
```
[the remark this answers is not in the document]
Claudenw: Changed

apache-rat-core/src/main/java/org/apache/rat/report/claim/impl/ClaimAggregator.java
Claudenw: ClaimAggregator historically tracks integer values.  However we have several cases where we have streams that we call `count()` on.  They return longs.  We could cast them or convert them here.  I have changed this to be an `int` value and cast the stream counts.  I don't think that we will have more than MAX_INT defined licenses categories or licenses.

apache-rat-core/src/main/java/org/apache/rat/report/claim/impl/xml/SimpleXmlClaimReporter.java
Claudenw: Stupid log error on my part.  Good catch.  We are indeed closing the element.

apache-rat-core/src/main/java/org/apache/rat/report/xml/writer/impl/base/XmlWriter.java
Claudenw: I think all overridden methods should rely on the interface for documentation. :(

apache-rat-core/src/main/java/org/apache/rat/report/xml/writer/impl/base/XmlWriter.java
Claudenw: It was supposed to be a real comment, though it is in the middle of a CDATA block.
Perhaps I should have inserted a new CDATA block start after the closure rather than removing the original cdata block.

The issure arises when samples from XML document include an ending CDATA tag.

apache-rat-core/src/main/java/org/apache/rat/utils/Log.java
Claudenw: `printStackTrace(writer)` will write to the writer and not to error.

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
Claudenw: Uppercase is now the default since the enums were changed.

apache-rat-core/src/test/java/org/apache/rat/ReportTest.java
```diff
@@ -55,9 +61,9 @@ public void testDefaultConfiguration() throws ParseException, IOException {
 
     @Test
     public void testOutputOption() throws Exception {
-        CommandLine cl = new DefaultParser().parse(Report.buildOptions(), new String[]{ "-o", "target/test" });
+        CommandLine cl = new DefaultParser().parse(Report.buildOptions(), new String[] { "-o", "target/test" });
         ReportConfiguration config = Report.createConfiguration("target/test-classes/elements", cl);
-        Reporter.report(config);
+        new Reporter(config).output();
         File output = new File("target/test");
```
ottlinger, line 67: Should we use jUnit Testfolders here instead?

apache-rat-core/src/test/java/org/apache/rat/analysis/matchers/AbstractMatcherTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.rat.analysis.IHeaderMatcher;
+import org.apache.rat.analysis.IHeaders;
+
+public class AbstractMatcherTest {
+
+    private IHeaders dummyHeader = makeHeaders(null, null);
+
+    protected void assertValues(IHeaderMatcher target, boolean... values) {
+        for (int i = 0; i < values.length; i++) {
+            final int idx = i;
```
ottlinger, line 34: do we need this assignment due to the lambda or can we use the counter/i instead?

Claudenw: assignment needed for the lambda.

apache-rat-core/src/test/java/org/apache/rat/ReportTest.java
Claudenw: Yes, fixed.

Claudenw:
Just a FYI: I have an implementation of a program that will write ANT tasks for any given command line configuration.  So adding new matchers in Ant is fairly easy.

apache-rat-tasks/pom.xml
```diff
@@ -105,7 +105,7 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
+      <!--<plugin>
```
ottlinger, line 108: Are these tests removed or did you change the testing mechanism for Ant?

apache-rat-tasks/src/test/resources/antunit/report-bad-configurations.xml
```diff
@@ -59,7 +58,7 @@
   </target>
 
   <target name="testNoStylesheet">
-    <rat:report format="styled">
+    <rat:report styleReport="styled">
```
ottlinger, line 61: Does that mean we need more docs about how to use RAT within the Ant sphere? Are these changes similar to the unintendedly breaking changes that were introduced with 0.16.0?
https://creadur.apache.org/rat/apache-rat-tasks/index.html
https://creadur.apache.org/rat/apache-rat-tasks/report.html

src/site/apt/index.apt.vm
```diff
@@ -63,6 +63,16 @@ Apache Rat
...
  {{{./mailing-lists.html}talk to us}}
  whenever Rat falls short.
+ 
+** How do I extend Rat
+ 
+ There are several standard ways to extend Rat.
+ 
+   * Add a {{{./license_def.html}license definition}} via an XML file.
+   
+   * Add a new {{{./matcher_def.html}Matcher definition}}. Requires Java expertice.
```
ottlinger, line 73: Typo: Java expertise

src/site/apt/index.apt.vm
```diff
@@ -63,6 +63,16 @@ Apache Rat
...
+   
+   * Add a new definition format.  Requires java expertice as well as expertice with the format.
```
ottlinger, line 75: Typo 2x: expertise;
I'd prefer to uppercase Java.

src/site/apt/license_def.apt.vm
```diff
@@ -0,0 +1,143 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+                   --------------------------
+                   How to define new licenses
+                   --------------------------
+
+How to define licenses in Apache Rat
+
+ All licenses in Apache Rat are defined in configuration files.  There is a default 
+ {{{https://github.com/apache/creadur-rat/blob/master/apache-rat-core/src/main/resources/org/apache/rat/default.xml}
```
ottlinger, line 25: Let us use ASF links instead:
https://gitbox.apache.org/repos/asf/creadur-rat/blob/master/apache-rat-core/src/main/resources/org/apache/rat/default.xml

src/site/apt/license_def.apt.vm
```diff
@@ -0,0 +1,143 @@
...
+ XML based configuration file}} that can server as a good example of various definitions.
```
ottlinger, line 26: Typo: that can serve

src/site/apt/license_def.apt.vm
```diff
@@ -0,0 +1,143 @@
...
+
+ It is possible to create new parsers for different configuration formats.  But that task is beyone the
```
ottlinger, line 28: Typo: is beyond the

src/site/apt/license_def.apt.vm
```diff
@@ -0,0 +1,143 @@
...
+
+ It is possible to create new parsers for different configuration formats.  But that task is beyone the 
+ scope of this document.  In this document we will address how to define licenses in the default XML format.
+ 
+ The XML document has a root node named "rat-config" which comprises 4 major sections: "Families", 
+ "Licenses", "Approved" and "Matchers".
+ 
+* Families
+ 
+ Families are groups that define licenses that share similarities.  Each family has an id and a name, and example
```
ottlinger, line 36: two sentences ? Each family has ..... An example .....

src/site/apt/license_def.apt.vm
```diff
@@ -0,0 +1,143 @@
...
++------------------------------------------+
+ 
+* Licenses
+
+ License elements have three properties: "family", "id", and "name". The family property must refer to
+ the "id" of a family element.  The family element can be defined in the current document or in another
+ included document.  It has to exist by the time all the configuration files are read.  The "id" element
+ is optional, if it is not provided the id will be set to the id of the associated family.  However, the 
+ "id" property must be unique across all licenses.  The "name" property is also optional.  If not specified
+ the name of the associated family will be used.  It is recommened that each license have a unique name
```
ottlinger, line 72: 2x typo: It is recommended that each license has a .....

src/site/apt/license_def.apt.vm
```diff
@@ -0,0 +1,143 @@
...
+    <approved>
+        <family license_ref='AL' />
+    </approved>
+ </rat-config>
++------------------------------------------+
+
+* Listing components
+
+ All the components (Licenses and Matchers) defined in the system can be displayed using the Documentation
+ tool.  To run the too execute:
```
ottlinger, line 131: Typo: tool

src/site/apt/license_def.apt.vm
```diff
@@ -0,0 +1,143 @@
...
+
++------------------------------------------+
+java -jar apache-rat-${project.version}.jar org.apache.rat.Documentation
++------------------------------------------+
+
+ The tool take the same arguments as the standard rat CLI, so additional license files can be defined.
```
ottlinger, line 137: Typo: Rat CLI

src/site/apt/license_def.apt.vm
```diff
@@ -0,0 +1,143 @@
...
+ </rat-config>
++------------------------------------------+
+
+* Listing components
+
+ All the components (Licenses and Matchers) defined in the system can be displayed using the Documentation
+ tool.  To run the too execute:
+
++------------------------------------------+
+java -jar apache-rat-${project.version}.jar org.apache.rat.Documentation
```
ottlinger, line 134: Would you mind using the same kind of linking as in 
https://creadur.apache.org/rat/apache-rat/index.html
"apache-rat/target/" ?

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+~~   limitations under the License.
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+                   --------------------------
+                   How to define new Matchers
+                   --------------------------
+
+How to define matchers in Apache Rat
+
+ Matchers in Apache Rat are paired with Builders.  A Matcher must implement the "IHeaderMatcher" interface and its associated Builder must implement the IHeaderMatcher.Builder interface.
```
ottlinger, line 24: Would it make sense to link to the classes within Gitbox? Maybe this helps in case of refactorings as it would yield a 404 .....

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+
+* A simple example
+
+** The Matcher implementation
+
+ For our example we will implement a Matcher that implements the phrase "Quality, speed and cost, pick any two” by looking for the occurrence of all three words anywhere in the header.  
+ In most cases is it simplest to extend the AbstractHeaderMatcher class as this class will handle setting of the unique ID for instances that do not otherwise have a unique id.
```
ottlinger, line 31: For matters of consistency: is there a difference between ID and id or Id?
I spotted the same in Javadoc of the changes and asked myself if there is a semantic difference between the two or if we should just decide to use one notation in all places.
Personally I'm fine with "Id" in the code and "id" in written text.

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+
++------------------------------------------+
+public interface IHeaderMatcher extends Component {
```
ottlinger, line 34: Not sure if there is a way to reference parts of sources, but in case it remains we should stick a big warning in IHeaderMatcher to keep the site up2date.

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+     * 
+     * @param headers the representations of the headers to check
+     * @return {@code true} if the matcher matches the text, {@code false} otherwise.
+     */
+    boolean matches(IHeaders headers);
+}
++------------------------------------------+
+
++------------------------------------------+
+public abstract class AbstractHeaderMatcher implements IHeaderMatcher {
```
ottlinger, line 65: Maybe it is sufficient to add plain references to both classes as you provide a full example afterwards ?!

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
++------------------------------------------+
+package com.example.ratMatcher;
+
+import org.apache.rat.analysis.IHeaders;
+import org.apache.rat.analysis.matchers.AbstractHeaderMatcher;
+import org.apache.rat.config.parameters.Component;
+import org.apache.rat.config.parameters.ConfigComponent;
+
+@ConfigComponent(type = Component.Type.Matcher, name = "QSC", desc = "Reports if the 'Quality, speed and cost, pick any two' rule is violated")
+public class QSCMatcher extends AbstraactHeaderMatcher {
```
ottlinger, line 108: Typo: extends AbstractHeaderMatcher

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+
+    public QSCMatcher(String id) {
+        super(id);
+    }
+
+    @Override
+    public boolean matches(IHeaders headers) {
+        String text = headers.prune()
+        return text.contains("quality") && text.contains("speed") && text.contains("cost");
```
ottlinger, line 117: Initially I thought of being able to use one of the Matchers provided within this PR - shouldn't I be able to "concatenate" matchers, such as ..... (pseudocode only):

AndMatcher.match(TextMatcher.contains("quality"), TextMatcher.contains("speed"), TextMatcher.contains("cost")) 

But maybe it is just too late now ;) and I misunderstood the example or above matchers in the code.

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+    }
+}
++------------------------------------------+
+
+ In the above example we use the ConfigComponent annotation to identify that this is a Matcher component, that it has the name 'QSC' and a description of what it does.  
+ If the "name" was not specified it would have been extracted from the class name by removing the "Matcher" from "QSCMatcher".
+
+ The Constructor calls the AbstractHeaderMatcher with a null argument so tha the parent class will assign an id.
```
ottlinger, line 125: typo: so that the

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+ lower case.  There is an issue with this matcher in that it would match the string: "The quality of Dennis Hopper's acting, as Keanu Reeves costar in 'Speed', is outstanding."
+ 
+ the correction of that is left as an exercise for the reader.  Hint: matching the pruned text can be a quick gating check for a set of more expensive regular expression checks. 
+
+** The Matcher.Builder implementation
+
+  The builder must implement the IHeaderMatcher.Builder interface.
+
++------------------------------------------+
+@FunctionalInterface
```
ottlinger, line 137: same as above I'd prefer to directly link to RAT sources/interfaces/classes via Gitbox instead of copy&pasting them here.

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+public class QSCBuilder extends AbstractBuilder. {
+    QSCMatcher build() {
+        return new QSCMatcher(getId());
+    }
+}
++------------------------------------------+
+
+** Registering the builder for use in XML configuration 
+
+  In order to use the matcher in a Rat configuration it has to be registered with the system.  This can be done by creating file with the builder specified and passing it
```
ottlinger, line 236: this can be done by creating a file? What kind of file? XML or configuration file?

ottlinger: Maybe the type just needs to be added for clarification purposes.

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+import org.apache.rat.analysis.IHeaders;
+import org.apache.rat.analysis.matchers.AbstractHeaderMatcher;
+import org.apache.rat.config.parameters.Component;
+import org.apache.rat.config.parameters.ConfigComponent;
+
+@ConfigComponent(type = Component.Type.Matcher, name = "TPM", desc = "Checks that the three string are found in the header")
+public class TPMatcher extends AbstraactHeaderMatcher {
```
ottlinger, line 275: Typo: AbstractHeaderMatcher

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+    @Override
+    public boolean matches(IHeaders headers) {
+        String text = headers.prune()
+        return text.contains(one) && text.contains(two) && text.contains(three);
+    }
+}
++------------------------------------------+
+
+ The ComponentConfigurations with the parameter type indicate that the members specify properties of the component.  The Matcher must have a "get" method for each parameter and the builder
+ must have a corresponding "set" method.  The names of the methods and the attributes in the XML paarser can be changed by adding a 'name' attribute to the ConfigComponent.
```
ottlinger, line 311: typo: parser

src/site/apt/matcher_def.apt.vm
```diff
@@ -0,0 +1,364 @@
...
+    <matchers>
+        <matcher class="com.example.ratMatcher.QSCBuilder" />
+        <matcher class="com.example.ratMatcher.TPBuilder" />
+    </matchers>
+</rat-config>
++------------------------------------------+
+
+* Embedded matchers.
+
+ It is possible to create matchers that embed other matchers.  The examples in the codebase are the "All", "Any" and "Not" matchers and their associated builders.
```
ottlinger, line 364: Could we link to the three matchers in Rat (Gitbox):
* All
* Any
* Not

I'd assume that anyone wanting to add new ones would like to take a closer look at these base matchers.

ottlinger:
@Claudenw thanks for these big changes with tests and documentation. It's a lot of code that changed, but the examples and tests seem much more simple and straightforward to me.

apache-rat-tasks/src/test/resources/antunit/report-bad-configurations.xml
Claudenw: format is deprecated (but still works), styleReport is the new attribute.

src/site/apt/license_def.apt.vm
Claudenw: "license have" is the future perfect tense (I think that is what it is called. "license has" would be past perfect.

src/site/apt/license_def.apt.vm
Claudenw: We should discuss this.  The index.html shows how to call the code if you have built it directly.  What we don't do is show how to run the tools if you have downloaded them.

I want to expand tools to include simple scripts to run the basic command line tools given a download of a compressed file, or something similar to that.  Make it easy to run for people who are just downloading and running.

src/site/apt/matcher_def.apt.vm
Claudenw: No semantic difference.  I have tried to clean the documentation to make it all "id"

src/site/apt/matcher_def.apt.vm
Claudenw: Currently we only have an XML configuration parser but the system is open to new implementations.  I have updated the documentation to reflect XML configuration.

src/site/apt/matcher_def.apt.vm
Claudenw: all links now point to ASF gitbox.

apache-rat-tasks/pom.xml
ottlinger: Thanks.

apache-rat-tasks/src/test/resources/antunit/report-bad-configurations.xml
ottlinger: I guess we should a not about the deprecation and the new attribute in:
https://github.com/apache/creadur-rat/blob/master/apache-rat-tasks/src/site/apt/report.apt

PR #238, opened 2024-04-16
RAT-369: Add checkstyle and spotbugs to build and generate a report

ottlinger:
I verified locally (via ./.buildtools/generateStagingSiteInWebpageRepo) that a report is generated, but the build is not yet configured to fail.


ottlinger:
@Claudenw in case of any errors the build fails. I'm not sure if it's that helpful or if we remove this setting on the branch .....

ottlinger:
In order to be able to bring changes intro master I'd prefer to configure limits per module with the current threshold, thus we could make sure that no new stuff is being added and can fix the errors&warning in an incremental way - WDYT @Claudenw ?

Claudenw:
My thought was to turn on the spotbugs and fix any critical errors at that time (the ones with the red circles). 
We should then open fixes to remove the yellow ones and/or determine if there are way to ignore them in specific cases (annotations perhaps?) so that we can fix the ones that need fixing and mark the ones that we determine are not an issue.

ottlinger:
@Claudenw the thing is that it either breaks the build or not. One can configure exclusions in an XML file in order to tolerate certain bugs explicitly ..... I'll have a look, but the first report said roughly up to 100 errors per submodule.

ottlinger:
Most of the errors relate to the fact that many file operations rely on the default charset. Unfortunately newer version of the JDK provide constructors to specify an encoding, but as RAT still relies on JDK8 these changes will be bigger.

apache-rat-tasks/src/main/java/org/apache/rat/anttasks/Report.java
```diff
@@ -318,7 +319,6 @@ public void log(Level level, String msg) {
                 write(Project.MSG_ERR, msg);
                 break;
 			case OFF:
-				break;
 			default:
 				break;
             }
```
Claudenw, line 324: It looks like the above lines are using tabs. can we convert them to spaces?

apache-rat-core/spotbugs_ignore.xml
```diff
@@ -0,0 +1,27 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<FindBugsFilter>
+  <Match>
+    <!-- Uses tryWithResources, thus closes resource properly:
```
Claudenw, line 20: I believe this can be fixed by changing:

```
            try (final InputStream in = url.openStream()) {
                BufferedReader buffer = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8));
```
to
```
            try (final InputStream in = url.openStream();
                BufferedReader buffer = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8))) {
```

apache-rat-plugin/spotbugs_ignore.xml
```diff
@@ -0,0 +1,27 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<FindBugsFilter>
+  <Match>
+    <!-- Uses tryWithResources, thus closes resource properly:
+    [ERROR] Medium: org.apache.rat.configuration.builders.ChildContainerBuilder.setResource(String) may fail to close stream
```
Claudenw, line 21: see previous spotbugs_ignore.xml comment

apache-rat-tasks/spotbugs_ignore.xml
```diff
@@ -0,0 +1,27 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<FindBugsFilter>
+  <Match>
+    <!-- Uses tryWithResources, thus closes resource properly:
+    [ERROR] Medium: org.apache.rat.configuration.builders.ChildContainerBuilder.setResource(String) may fail to close stream
```
Claudenw, line 21: see previous spotbugs_ignore.xml comment

apache-rat/spotbugs_ignore.xml
```diff
@@ -0,0 +1,27 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<FindBugsFilter>
+  <Match>
+    <!-- Uses tryWithResources, thus closes resource properly:
+    [ERROR] Medium: org.apache.rat.configuration.builders.ChildContainerBuilder.setResource(String) may fail to close stream
```
Claudenw, line 21: see previous spotbugs_ignore.xml comment

spotbugs_ignore.xml
```diff
@@ -0,0 +1,27 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<FindBugsFilter>
+  <Match>
+    <!-- Uses tryWithResources, thus closes resource properly:
+    [ERROR] Medium: org.apache.rat.configuration.builders.ChildContainerBuilder.setResource(String) may fail to close stream
```
Claudenw, line 21: see previous spotbugs_ignore.xml comment

Claudenw (changes requested):
Looks mostly good to me.  Just 2 real changes.

apache-rat-core/spotbugs_ignore.xml
ottlinger: The problem is that these new convenience constructors were added after JDK8 and break the build that checks for JDK8-compliance. Thus I'm not sure how to fix these inherent "relying on default/system encoding" parts of RAT

apache-rat-plugin/spotbugs_ignore.xml
ottlinger: I had to add an exclusion file in each submodule to not break the build. Each one has an example entry .... should I try with empty exclusion files or mark it as an example in all submodules? @Claudenw

apache-rat-tasks/src/main/java/org/apache/rat/anttasks/Report.java
ottlinger: Done - thanks.

apache-rat-core/spotbugs_ignore.xml
ottlinger: Sorry I checked again - this particular example is fixed as you suggested. Should I remove it from the exclusion files?

apache-rat-plugin/src/main/java/org/apache/rat/mp/util/ignore/GlobIgnoreMatcher.java
```diff
@@ -25,11 +25,7 @@
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
```
Claudenw, line 28: Do not use '*" format for includes.  If you are using IntelliJ you can specify a large number of imports in the same package before * to stop this.

apache-rat-core/src/main/java/org/apache/rat/config/parameters/ComponentType.java
```diff
@@ -28,6 +28,6 @@ public enum ComponentType {
     MATCHER,
     /** A Parameter for example the "id" parameter found in every component */
     PARAMETER,
-    /** A parameter that is supplied by the environment.  Currently systems using builders have to handle seting this.  For example the list of matchers for the "MatcherRefBuilder" */
-    BULID_PARAMETER 
+    /** A parameter that is supplied by the environment.  Currently systems using builders have to handle setting this.  For example the list of matchers for the "MatcherRefBuilder" */
+    BUILD_PARAMETER
```
Claudenw, line 32: Thanks for catching this.

Claudenw (changes requested):
Mostly looks good.  import * is the only real problem

ottlinger:
Could we merge this branch and fix the remaining errors step-by-step? Personally I'd prefer to have shorter-living branches ;) @Claudenw

apache-rat-plugin/src/main/java/org/apache/rat/mp/util/ignore/GlobIgnoreMatcher.java
ottlinger: Fixed.

spotbugs_ignore.xml
```diff
@@ -0,0 +1,26 @@
...
+    [org.apache.rat.configuration.builders.ChildContainerBuilder] At ChildContainerBuilder.java:[line 62] OS_OPEN_STREAM
+    <Class name="org.apache.rat.configuration.builders.ChildContainerBuilder"/>
+    <Bug pattern="OS_OPEN_STREAM"/-->
+  </Match>
```
Claudenw, line 25: Should the commented code above be removed.  If so should the file be removed?

Claudenw (approved):
Looks good. though I don't see any report of spotbugs being generated.  I was hoping we could get something that would tell us if the number of bugs was decreasing over time.

spotbugs_ignore.xml
ottlinger: If the file is missing the build fails, therefore I had to introduce it in all submodules.

ottlinger: The maven spotbugs integration allows either to run spotbugs:spotbugs to see the results and prints its results into the build log. Apart from that it is integrated into the site build so that each submodule has a spotbugs-report.

Spotbugs allows to configure a threshold ... I'll see if I can integrate this to at least fail the builds once a new error is introduced:
https://spotbugs.github.io/spotbugs-maven-plugin/spotbugs-mojo.html

ottlinger: Unfortunately this is not what we would want to have :( The docs are not really that good as no description is provided.

ottlinger: The only available exclusion rule seems to relate to the severity ....
https://github.com/spotbugs/spotbugs-maven-plugin/blob/master/src/main/groovy/org/codehaus/mojo/spotbugs/BaseViolationCheckMojo.groovy

ottlinger: I assume mvn spotbugs:gui is the best way to work with the tool at the moment as we generate the webpage manually before an upcoming release .....

ottlinger:
@Claudenw If you do not mind I would merge this one tomorrow :) Thanks for your feedback.

Claudenw:
Sounds good to me

ottlinger:
@Claudenw I filed https://github.com/spotbugs/spotbugs-maven-plugin/issues/804 to maybe get some feedback whether we can use spotbugs in a different way to allow failing the build if new errors are introduced.

PR #240, opened 2024-04-18
RAT-54: Tika based document analyzer

Claudenw:
Switches to Tika to determine document type.

** this change depends upon #233 **
After the above change is merged this change will be approx 10 files -- most of them tests.

json files are now STANDARD not BINARY

fixes RAT-20  Detection of binaries should be smarter
fixes RAT-54  MIME Detection Using Tika
fixes RAT-147 binary guesser design improvement
fixes RAT-150 RAT should use Apache Tika to simply guess ignored [application/X] file types and focus on the [text/Y] family as a sensible default
fixes RAT-211 Generated rat-output.xml must be well-formed, even if BinaryGuesser fails
fixes RAT-301 Chinese characters comments are not recognized as binary anymore (thanks to Tika)

Checklist

- [X] add documentation for default file exclusion listing *.json files as the excluded file.  Also note that setting the exclusion will replace the existing entry, so it should be included in replacements. **NOT NEEDED** as .json is no longer excluded but again categorized as BINARY.
- [x] add documentation for the default directory exclusion of hidden (.*) files.  Also note that setting the exclusion will replace the existing entry.  (There is no non-programatic access to changing the directory)
- [X] Add  tests for ArchiveWalker.  (https://issues.apache.org/jira/browse/RAT-372)


ottlinger:
I'd prefer a file filter that allows ignoring no-comment plaintext files such as JSON.

apache-rat-core/pom.xml
```diff
@@ -126,5 +126,10 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
ottlinger, line 129: Could we add the dependency into a dependencyManagement block to only have the version in 1 pom.xml and reference it in the other one?

apache-rat-core/src/main/java/org/apache/rat/api/Document.java
```diff
@@ -33,47 +36,416 @@ public interface Document {
...
+                return UNKNOWN;
+            }
+            return result;
+        }
+
+        /*
+         * https://tika.apache.org/3.0.0-BETA/formats.html 
+         */
+        static {
+            documentTypeMap = new HashMap<>();
```
ottlinger, line 66: Seeing this long list I wondered if we should add a new module apache-rat-regression-tests that contains at least one example for all the file types and can be used measure and integration-test RAT .....

Claudenw: Shall we just open a JIRA ticket for this.  I think that collecting some of the files will be difficult.

ottlinger: I created RAT-370 for that.

Claudenw:
I am adding a  file filter to remove json files.  Initially this will be hard coded.  I will open a subsequent ticket to generalize it so that we can define a list of extensions to remove.  This will probably mean more command line options.  So I'll have to open that can of worms as well.

Claudenw:
Well this blew up to something bigger than I wanted but...

I added default exclusion for "*.json" files in the Default class and used that to configure the ReportConfiguration defaults.
In the future the default exclusions should be configured in Default class.

I changed the parameter names and instance variables to "filesToIgnore" and "directoriesToIgnore" to make it clear what the filters were doing.

I cleaned up the DirectoryWalker and Walker classes.

I ensured that the filesToIgnore and directoriesToIgnore alwasy have a value (not null).  If set to null in the configuration the value is translated into a filter that always returns false.

I think there may be an issue in the directoreisToIgnore processing, but it has been there from the beginning.  I will investigate and open a ticket if necessary.

apache-rat-core/src/main/java/org/apache/rat/Defaults.java
```diff
@@ -57,6 +62,10 @@ public class Defaults {
     public static final String UNAPPROVED_LICENSES_STYLESHEET = "org/apache/rat/unapproved-licenses.xsl";
 
     private final LicenseSetFactory setFactory;
+
+    private final FilenameFilter filesToIgnore = WildcardFileFilter.builder().setWildcards("*.json").setIoCase(IOCase.INSENSITIVE).get();
+
+    private final IOFileFilter directoriesToIgnore = NameBasedHiddenFileFilter.HIDDEN;
```
ottlinger, line 68: Would it make sense to add these 2 special cases to the documentation?
Didn't you (or JB) add a configuration option to scan for hidden files?

ottlinger: Should we create a new class for these 2 members? ScanDefaults that contains files and directories to ignore?

ScanDefault {
List<String> filesToIgnore; // wildcards
List<IOFileFilter> directoriesToIgnore; // not sure if IOFileFilter is the correct superclass
}

WDYT?

This would allow to have a static version of this configuration set in Defaults.java such as

public static ScanDefault RAT_DEFAULT_SCAN = new ScanDefault(List.of(*.json),List.of(NamebasedHiffenFilterFilter.HIDDEN);

apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -452,11 +452,11 @@ private static IReportable getDirectory(String baseDirectory, ReportConfiguratio
             }
 
             if (base.isDirectory()) {
-                return new DirectoryWalker(base, config.getInputFileFilter(), config.getDirectoryFilter());
+                return new DirectoryWalker(base, config.getFilesToIgnore(), config.getDirectoriesToIgnore());
```
ottlinger, line 455: DirectoryWalker would also benefit from a separate class ScanDefault .... WDYT?

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -179,31 +177,31 @@ public boolean isDryRun() {
...
-    public IOFileFilter getDirectoryFilter() {
-        return directoryFilter;
+    public IOFileFilter getDirectoriesToIgnore() {
+        return directoriesToIgnore;
     }
 
-    public void setDirectoryFilter(IOFileFilter directoryFilter) {
-        if (directoryFilter == null) {
-            this.directoryFilter = FalseFileFilter.FALSE;
+    public void setDirectoriesToIgnore(IOFileFilter directoriesToIgnore) {
```
ottlinger, line 195: ScanDefault could have boolean hasFiles()/hasDirectories() methods to not have to handle null values while interacting with the scan configuration

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
```diff
@@ -57,45 +58,71 @@ public int getCounter(Counter counter) {
...
-     */
-    public Map<Counter, int[]> getCounterMap() {
-        return counterMap;
+    public void incCounter(Counter key, int value) {
+        final int[] num = counterMap.get(key);
+
+        if (num == null) {
+            counterMap.put(key, new int[] { value });
+        } else {
+            num[0] += value;
```
ottlinger, line 67: are we running into problems here, when we do multithreaded analysis? Do we need AtomicInteger here?

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
```diff
@@ -57,45 +58,71 @@ public int getCounter(Counter counter) {
...
-     */
-    public Map<String, int[]> getLicenseFamilyCodeMap() {
-        return licenseFamilyCodeMap;
+    public void incCounter(Document.Type documentType, int value) {
+        final int[] num = documentCategoryMap.get(documentType);
+
+        if (num == null) {
+            documentCategoryMap.put(documentType, new int[] { value });
+        } else {
+            num[0] += value;
```
ottlinger, line 87: AtomicInteger to be threadsafe?

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
```diff
@@ -57,45 +58,71 @@ public int getCounter(Counter counter) {
...
+        return count == null ? 0 : count[0];
+    }
+
+    public void incLicenseFamilyCount(String licenseFamilyName, int value) {
+        final int[] num = licenseFamilyCodeMap.get(licenseFamilyName);
+
+        if (num == null) {
+            licenseFamilyCodeMap.put(licenseFamilyName, new int[] { value });
+        } else {
+            num[0] += value;
```
ottlinger, line 102: AtomicInteger?

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
```diff
@@ -57,45 +58,71 @@ public int getCounter(Counter counter) {
...
+        return count == null ? 0 : count[0];
+    }
+
+    public void incLicenseFileNameCount(String licenseFileNameName, int value) {
+        final int[] num = licenseFamilyNameMap.get(licenseFileNameName);
+
+        if (num == null) {
+            licenseFamilyNameMap.put(licenseFileNameName, new int[] { value });
+        } else {
+            num[0] += value;
```
ottlinger, line 125: AtomicInteger?

apache-rat-core/src/main/java/org/apache/rat/walker/DirectoryWalker.java
```diff
@@ -38,56 +38,42 @@ public class DirectoryWalker extends Walker implements IReportable {
...
+     * @param filesToIgnore filters input files (optional),
      *               or null when no filtering should be performed
-     * @param directoryFilter filters directories (optional), or null when no filtering should be performed.
+     * @param directoriesToIgnore filters directories (optional), or null when no filtering should be performed.
      */
-    public DirectoryWalker(File file, final FilenameFilter filter, IOFileFilter directoryFilter) {
-        super(file.getPath(), file, filter);
-        this.directoryFilter = directoryFilter;
+    public DirectoryWalker(File file, final FilenameFilter filesToIgnore, IOFileFilter directoriesToIgnore) {
+        super(file.getPath(), file, filesToIgnore);
```
ottlinger, line 52: Maybe this class would also benefit from ScanDefault as mentioned in above comments

apache-rat-core/src/main/java/org/apache/rat/walker/Walker.java
```diff
@@ -33,38 +34,32 @@ public abstract class Walker implements IReportable {
...
 
     protected static FilenameFilter regexFilter(final Pattern pattern) {
         return (dir, name) -> {
             final boolean result;
             if (pattern == null) {
-                result = true;
+                result = false;
             } else {
-                result = !pattern.matcher(name).matches();
+                result = pattern.matcher(name).matches();
```
ottlinger, line 45: Thanks for getting rid of the negation. Has there been an error in here before or should line 42 be changed as well?

BEFORE
if(pattern==null) return true
else 
!pattern.matches()

NOW
if(pattern==null) return false
else 
pattern.matches()

or did I not get what is going on here?

src/changes/changes.xml
```diff
@@ -72,6 +72,22 @@ https://maven.apache.org/plugins/maven-changes-plugin/xsd/changes-1.0.0.xsd
...
     -->
     <release version="0.17-SNAPSHOT" date="xxxx-yy-zz" description="Current SNAPSHOT - release to be done">
+      <action issue="RAT-54" type="fix" dev="claudenw">
+        MIME Detection Using Tika
+      </action>
+      <action issue="RAT-20" type="fix" dev="claudenw">
+        Changed to detecting binary by content not name.
+      </action>
+      <action issue="RAT-147" type="fix" dev="claudenw">
+        Change to detect non UTF-8 text as text not binary.
```
ottlinger, line 82: Isn't that what RAT-301 is all about? I got curious when I read your comment in the changelog here.

ottlinger (commented):
Thanks for all the fixes while introducing Tika to RAT - big kudos.

Should we add a note to the webpage or is it just an entry in the release changelog concerning the use of Tika for file type detection? WDYT?

apache-rat-core/src/main/java/org/apache/rat/Defaults.java
Claudenw: Defaults is intended to be the System defaults for the ReportConfiguration.  There are some cases where the report option has to be set by the UI before the Defaults can be tested.  And there is a flag for no defaults so the Defaults need to be specified outside of the ReportConfiguration initialization.  

I changed the description of Defaults and updated the values and methods to be static.

apache-rat-core/src/main/java/org/apache/rat/Report.java
Claudenw: I  think that the ReportConfiguration class needs to be reworked into enclosed objects.  The enclosed objects would be something like ScanDefault.  However, I also think that the definition of the objects needs to be based on the structure of the command line client, and that the command line client needs to be reworked to have better names for the options that signify what part of the config is being set.  The names are currently too short to be meaningful.

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
Claudenw: Report configuration is the only place where this is handled so that everyplace we want the directoriesToIgnore it is set and there is no need to do the null check.  This case allows UI to set directoriesToIgnore to null (the old setting) and still guarantee the directoriesToIgnore has a value.  In future we might change the UIs to require that they send a defined static value to signal no directories.  But this is the fix without modifying the UIs.

This also applies for filesToIgnore.

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
Claudenw: I had not thought about atomic properties, but I did think the array solution is no longer required.  I think I have a solution using a concurrent map and a IncrementableInt class.  I will put that together shortly.

apache-rat-core/src/main/java/org/apache/rat/walker/Walker.java
Claudenw: I will go back and double check that this correct.  However, I did change the nomenclature to be speicfic, the list of  `filesToIgnore` should indicate that if the pattern is matched the file is excluded (same with directory).  Previously the name was `fileFilter`  and there it is difficult to know if you are including or excluding files in the filter.

src/changes/changes.xml
Claudenw: 147 is improve the guesser so that non UTF-8 text is not detected as binary.
301 is extended (Chinese in the report) characters that are UTF-8 encoded being detected as binary.

related but not the same.  I was waiting for example to test 301 with.

ottlinger: https://github.com/apache/linkis/blob/master/linkis-public-enhancements/linkis-pes-common/src/main/java/org/apache/linkis/udf/entity/UDFVersion.java

Claudenw:
I extracted the Tika processing to its own class.  
I added the tika `MediaType` to our metadata.
The process now assumes that all media types = "text/*" are `STANDARD` documents, 
for everything else it is `BINARY` unless it is listed in the `documentTypeMap` which is now in the `TikaProcessor` class.

You will notice that `application/json` is listed in the `documentTypeMap`.  This is a stupid move on my part and will be removed.  This will mean that we can then remove the `WildcardFileName` filter for *.json from the `filesToIgnore` filter in `ReportConfiguration`.  I will fix this soon.

apache-rat-core/src/main/java/org/apache/rat/Defaults.java
```diff
@@ -152,27 +163,35 @@ public SortedSet<ILicenseFamily> getLicenseFamilies(LicenseFilter filter) {
...
+    public final static class Builder {
         private final Set<URL> fileNames = new TreeSet<>(Comparator.comparing(URL::toString));
 
         private Builder() {
             fileNames.add(DEFAULT_CONFIG_URL);
         }
 
         /**
-         * Adds a URL to a configuration file to be read.
+         * Adds a URI to a configuration file to be read.
```
ottlinger, line 189: URI or URL? Spotbugs complains that URL-comparison may lead to performance issues as each URL is resolved/queried via DNS ..... simply changing to URI is only psssible with more changes within RAT.

apache-rat-core/src/main/java/org/apache/rat/analysis/DefaultAnalyserFactory.java
```diff
@@ -63,8 +60,8 @@ private final static class DefaultAnalyser implements IDocumentAnalyser {
 
         /**
          * Constructs a DocumentAnalyser for the specified license.
-         * 
-         * @param license The license to analyse
+         * @param log the Log to use
+         * @param licenses The license to analyse
```
ottlinger, line 64: typo: licenses to analyse

apache-rat-core/src/main/java/org/apache/rat/Defaults.java
Claudenw: Agree, I started to look at it but I think we need a ticket just to handle the URL -> URI conversion.

Claudenw: I opened ticket https://issues.apache.org/jira/browse/RAT-371 to deal with URL -> URI transition.

ottlinger: Thanks.

ottlinger:
@Claudenw the extraction into the Tika-class looks very nice - thanks.

Claudenw:
@ottlinger If you approve I can merge this.  If you want more eyes on it, lets's invite a few reviewers.

apache-rat-core/pom.xml
ottlinger: Thanks.

apache-rat-core/src/test/java/org/apache/rat/document/impl/guesser/BinaryGuesserTest.java
```diff
@@ -1,150 +1,150 @@
...
+//import java.io.IOException;
+//import java.io.Reader;
+//import java.util.Arrays;
+//import java.util.List;
+//
+//import static org.junit.jupiter.api.Assertions.assertEquals;
+//import static org.junit.jupiter.api.Assertions.assertFalse;
+//import static org.junit.jupiter.api.Assertions.assertTrue;
+//
+//public class BinaryGuesserTest {
```
ottlinger, line 36: Should we enable and adapt the test for binaries?

src/changes/changes.xml
ottlinger: @Claudenw would you mind integrating above file in order to see if its properly detected with Tika?

ottlinger:
> @ottlinger If you approve I can merge this. If you want more eyes on it, lets's invite a few reviewers.

In the PR's main description you've created a check list - is that already done? 
I'm still thinking if we should add more documentation about the change or leave to a bigger entry in the release notes about possibly changing behaviour of RAT.

Claudenw:
I updated the checklist.

src/changes/changes.xml
ottlinger: @Claudenw I brought in changes related to RAT-301 - is that fine for you or should the Chinese character example go somewhere else? Thanks

ottlinger:
@Claudenw pls review my latest additions concerning RAT-301, after that go ahead with the merge. Thanks for your work and the cool addition of more functionality to RAT #kudos

ottlinger (approved):
Thanks for all your work.

PR #242, opened 2024-04-23
RAT-355, RAT-366: Fixed closed output issue

Claudenw:
This fixes an issue detected in the build output where the tail end of the output disappears, though the build appears to work correctly.

The issue is that the ant report test manages to close System.out.  This change ensures that System.out does not get closed.


ottlinger (approved):
Thanks.

ottlinger:
Confirmed locally - the build looks okay again.

PR #246, opened 2024-05-04
RAT-372: Create archive processor

Claudenw:
Initial implementation 
see https://issues.apache.org/jira/browse/RAT-372
See discussion:  https://lists.apache.org/thread/gyp2tfsw3w66x0wy95tv1bcwb667h866

## Checklist

- [x] This pull is built upon the Tika changes in RAT-54 (#240) and should not be considered for merge until and unless that is merged.
- [x] Significant documentation needs to be put in place to discuss how the `--archive` option affects output.
- [x] Add documentation for XSLT building.
- [x] Add Changelog with all possible changes introduced with that PR
- [x] Add changelog / Notice entries for deprecated CLI options.


ottlinger:
Let's review this after https://github.com/apache/creadur-rat/pull/240 is merged into master - thanks.

Claudenw:
This change also updated the Help routines and the CLI usage of commons-cli.  New help screen now looks like:

```
usage: java -jar apache-rat/target/apache-rat-CURRENT-VERSION.jar [options] [DIR|TARBALL]

====== Available Options ======
 -a                                   (deprecated) Add the default license header to any file with an unknown license.  Use '-A'
                                      or ---addLicense instead.
 -A,--addLicense                      Add the default license header to any file with an unknown license that is not in the
                                      exclusion list. By default new files will be created with the license header, to force the
                                      modification of existing files use the --force option.
    --archive <ProcessingType>        Specifies how ARCHIVE processing will be handled. (default is NOTIFICATION)
 -c,--copyright <arg>                 The copyright message to use in the license headers, usually in the form of "Copyright 2008
                                      Foo"
 -d,--dir <DirOrArchive>              (deprecated, use '--') Used to indicate source when using --exclude.
    --dry-run                         If set do not update the files but generate the reports.
 -e,--exclude <Expression>            Excludes files matching wildcard <expression>. May be followed by multiple argumentsNote
                                      that '--' or a following option is required when using this parameter.
 -E,--exclude-file <FileOrURI>        Excludes files matching regular expression in the input file.
 -f,--force                           Forces any changes in files to be written directly to the source files (i.e. new files are
                                      not created).
 -h,--help                            Print help for the RAT command line interface and exit.
    --licenses <FileOrURI>            File names or URLs for license definitions
    --list-families <LicenseFilter>   List the defined license families (default is NONE). Valid options are: ALL, APPROVED, NONE
    --list-licenses <LicenseFilter>   List the defined licenses (default is NONE). Valid options are: ALL, APPROVED, NONE
    --log-level <LogLevel>            sets the log level.
    --no-default-licenses             Ignore default configuration. By default all approved default licenses are used
 -o,--out <arg>                       Define the output file where to write a report to (default is System.out).
 -s,--stylesheet <arg>                XSLT stylesheet to use when creating the report.  Not compatible with -x.  Either an
                                      external xsl file may be specified or one of the internal named sheets: plain-rat (default),
                                      missing-headers, or unapproved-licenses
    --scan-hidden-directories         Scan hidden directories
 -x,--xml                             Output the report in raw XML format.  Not compatible with -s

====== Argument Types ======

<DirOrArchive>
      A director or archive file to scan

<Expression>
      A wildcard file matching pattern. example: *-test-*.txt

<FileOrURI>
      A file name or URI

<LicenseFilter>
      A defined filter for the licenses to include.  Valid values: ALL, APPROVED, NONE.

<LogLevel>
      The log level to use.  Valid values DEBUG, INFO, WARN, ERROR, OFF.

<ProcessingType>
      Specifies how to process file types.  Valid values are:
           NOTIFICATION: List archive as present
           PRESENCE: List any licenses found
           ABSENCE: List licenses found and any unknown licences

<StyleSheet>
      Either an external xsl file may be one of the internal named sheets: missing-headers, plain-rat, unapproved-licenses.

====== Notes ======

1. Rat highlights possible issues.
2. Rat reports require interpretation.
3. Rat often requires some tuning before it runs well against a project.
4. Rat relies on heuristics: it may miss issues

```

apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -59,74 +65,166 @@
...
 public class Report {
+
+    private static final String[] NOTES = {
+            "Rat highlights possible issues.",
+            "Rat reports require interpretation.",
+            "Rat often requires some tuning before it runs well against a project.",
+            "Rat relies on heuristics: it may miss issues"
+    };
+
+    private static final String[] STYLE_SHEETS = {"plain-rat", "missing-headers", "unapproved-licenses"};
```
ottlinger, line 82: Should we extract this into an enum - 
apart from the text/CLI option value it could also contain the location of the corresponding XSLT and a description.

WDYT?

Claudenw: We could do so and provide a descriptoin.  We don't need the location as using `-s missing-headers` is all that is needed to specify the internal XSLT files.  We  might note where they are in comments so that developers can find them easily.  Or point to them from documentation for XSLT development.

Claudenw: I made the change to an enum in the latest push.  I also updated the site documentation to give background on how to write the XSLT.

Claudenw:
For an interesting test run the following;
```
mvn clean install
java -jar apache-rat/target/apache-rat-0.17-SNAPSHOT.jar --archive PRESENCE .
```

Claudenw:
Another interesting run

```
mvn clean install
mvn dependency:copy-dependencies -DoutputDirectory=/tmp/lib
java -jar apache-rat/target/apache-rat-0.17-SNAPSHOT.jar --archive PRESENCE /tmp/lib
```

apache-rat-core/src/main/java/org/apache/rat/document/impl/FileDocument.java
```diff
@@ -19,67 +19,71 @@
...
-    public boolean isComposite() {
-        return DocumentImplUtils.isZip(file);
+    /**
+     * Normalizes a file name.  Accounts for Windows to Unix conversion.
+     * @param file
+     * @return
+     */
+    public final static String normalizeFileName(File file) {
+        String path = file.getPath();
+        return path.replace('\\', '/');
```
ottlinger, line 59: Does this destroy escaped filenames with spaces in their file names? Just wondering

src/changes/changes.xml
```diff
@@ -107,12 +119,15 @@ https://maven.apache.org/plugins/maven-changes-plugin/xsd/changes-1.0.0.xsd
...
+
+        This change also resulted in the possibility of multiple licenses being detected and reported.  forcing a change in the
+        XML ouptut.  XML schema was developed for the output.
       </action>
       <action issue="RAT-333" type="fix" dev="claudenw">
         Fix if --force option is used executable bit is not set properly on newly created/license-augmented file.
       </action>
       <action issue="RAT-345" type="update" dev="pottlinger" due-to="Niels Basjes">
-        Update gitignore-reader from 1.4.0 to 1.5.1 to fetch changes resulting from fixes of RAT-362.
+        Update gitignore-reader from 1.4.0 to 1.5.1 to fetch changes resulting from fixes of RAT-362.
```
ottlinger, line 130: What is the change here? as these entries are from older relases, right?

apache-rat-core/src/main/java/org/apache/rat/document/impl/FileDocument.java
Claudenw: To be honest, I don't know.  However, it is the same code that was found in a static class earlier.  I just moved it here modified.  I changed the name from `asName(File)` (?) to `normalizeFileName(File)`

src/changes/changes.xml
Claudenw: The entries had some weird [NBSP] (non break space?) character, I replaced them with a space.  I don't think it matters, but it made my IntelliJ editor much happier.

Claudenw:
The you can adjust the chatty level with the `--archive` settings.  By default it is no more chatty than the earlier code.

I did change the XSLT so that the amount of space taken by the archive listing is reduced by 1/2, but if you ask for archive data to be added to the report, it has to get more chatty.

Is there a specific change you can recommend?

apache-rat-core/src/main/java/org/apache/rat/document/impl/FileDocument.java
ottlinger: Let's wait for any user feedback. Thanks.

src/changes/changes.xml
ottlinger: ok, go ahead :)

ottlinger:
> The you can adjust the chatty level with the `--archive` settings. By default it is no more chatty than the earlier code.
> 
> I did change the XSLT so that the amount of space taken by the archive listing is reduced by 1/2, but if you ask for archive data to be added to the report, it has to get more chatty.
> 
> Is there a specific change you can recommend?

Not really, it basically was more or less a gut feeling when I saw the new report in a play project of mine.

PR #249, opened 2024-05-09
RAT-377: Create a STANDARD process filter like the  ARCHIVE process switch in RAT-372

Claudenw:
Fixes RAT-377

This change is dependent upon RAT-372 and should not be considered for merged until after that change has been accepted.

The solution for process archive files (RAT-372) introduces a command line switch to change from just reporting the presence of archive files to reporting what licenses they contain or reporting licenses contains as well as missing licenses.

This request is for a similar change in STANDARD reporting.

The use case is for non-ASF projects that don't require a license header in every file but do require licensing to be declared within the project.

Using `--standard PRESENCE`  would cause Rat to list all the declared licenses in the scan, without the overhead noise of 'missing license headers'

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/Report.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/Report.java
@@ -252,8 +252,17 @@ public class Report {
      * @since 0.17.0
      */
     static final Option ARCHIVE = Option.builder().longOpt("archive").hasArg().argName("ProcessingType")
-            .desc(format("Specifies the level of detail in ARCHIVE reporting. (default is %s)",
+            .desc(format("Specifies the level of detail in ARCHIVE file reporting. (default is %s)",
                     ReportConfiguration.Processing.NOTIFICATION))
+            .converter(s -> ReportConfiguration.Processing.valueOf(s.toUpperCase()))
+            .build();
+
+    /**
+     * Specify the processing of STANDARD files.
+     */
+    static final Option STANDARD = Option.builder().longOpt("standard").hasArg().argName("ProcessingType")
+            .desc(format("Specifies the level of detail in STANDARD file reporting.. (default is %s)",
+                    Defaults.STANDARD_PROCESSING))
             .converter(s -> ReportConfiguration.Processing.valueOf(s.toUpperCase()))
             .build();
 
@@ -367,6 +376,14 @@ public class Report {
             }
         }
 
+        if (cl.hasOption(STANDARD)) {
+            try {
+                configuration.setStandardProcessing(cl.getParsedOptionValue(STANDARD));
+            } catch (ParseException e) {
+                logParseException(e, STANDARD, cl, Defaults.STANDARD_PROCESSING);
+            }
+        }
+
         if (cl.hasOption(OUT)) {
             try {
                 configuration.setOut((File) cl.getParsedOptionValue(OUT));
@@ -411,7 +428,6 @@ public class Report {
                 }
 
                 URL url = Report.class.getClassLoader().getResource(String.format("org/apache/rat/%s.xsl", style[0]));
-
                 IOSupplier<InputStream> ioSupplier = (url == null) ?
                         () -> Files.newInputStream(Paths.get(style[0])) :
                         url::openStream;
@@ -470,6 +486,7 @@ public class Report {
     static Options buildOptions() {
         return new Options()
                 .addOption(ARCHIVE)
+                .addOption(STANDARD)
                 .addOption(DRY_RUN)
                 .addOption(LIST_FAMILIES)
                 .addOption(LIST_LICENSES)
--- a/apache-rat-core/src/main/java/org/apache/rat/Report.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/Report.java
@@ -261,7 +261,7 @@ public class Report {
      * Specify the processing of STANDARD files.
      */
     static final Option STANDARD = Option.builder().longOpt("standard").hasArg().argName("ProcessingType")
-            .desc(format("Specifies the level of detail in STANDARD file reporting.. (default is %s)",
+            .desc(format("Specifies the level of detail in STANDARD file reporting. (default is %s)",
                     Defaults.STANDARD_PROCESSING))
             .converter(s -> ReportConfiguration.Processing.valueOf(s.toUpperCase()))
             .build();
--- a/apache-rat-core/src/main/java/org/apache/rat/Report.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/Report.java
@@ -531,7 +531,7 @@ public class Report {
 
         String argumentPadding = createPadding(helpFormatter.getLeftPadding() + 5);
         for (Map.Entry<String, Supplier<String>> argInfo : ARGUMENT_TYPES.entrySet()) {
-            writer.format("\n<%s>\n", argInfo.getKey());
+            writer.format("%n<%s>%n", argInfo.getKey());
             helpFormatter.printWrapped(writer, helpFormatter.getWidth(), helpFormatter.getLeftPadding() + 10,
                     argumentPadding + argInfo.getValue().get());
         }
--- a/apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
@@ -58,7 +58,13 @@ import org.apache.rat.utils.ReportingSet;
  */
 public class ReportConfiguration {
 
-    public enum Processing {NOTIFICATION("List file as present"), PRESENCE("List any licenses found"), ABSENCE("List licenses found and any unknown licences");
+    public enum Processing {
+        /** List file as present only */
+        NOTIFICATION("List file as present"),
+        /** List all present licenses */
+        PRESENCE("List any licenses found"),
+        /** List all present licenses and unknown licenses */
+        ABSENCE("List licenses found and any unknown licences");
 
     private final String description;
     Processing(String description) {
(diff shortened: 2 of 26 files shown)


apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -246,11 +246,20 @@ public String desc() {
...
+            .desc(format("Specifies the level of detail in ARCHIVE file reporting. (default is %s)",
                     ReportConfiguration.Processing.NOTIFICATION))
             .converter(s -> ReportConfiguration.Processing.valueOf(s.toUpperCase()))
             .build();
 
+    /**
+     * Specify the processing of STANDARD files.
+     */
+    static final Option STANDARD = Option.builder().longOpt("standard").hasArg().argName("ProcessingType")
+            .desc(format("Specifies the level of detail in STANDARD file reporting.. (default is %s)",
```
ottlinger, line 258: Double dot in this sentence - pls update the site template as well to keep it in sync with your changes.

apache-rat-core/src/main/java/org/apache/rat/Report.java
```diff
@@ -405,7 +422,6 @@ static ReportConfiguration createConfiguration(String baseDirectory, CommandLine
                 }
 
                 URL url = Report.class.getClassLoader().getResource(String.format("org/apache/rat/%s.xsl", style[0]));
-
```
ottlinger, line 408: Should we add a tryCatch around this trying to load the xsl file in case user input is invalid/not found?

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -121,6 +127,22 @@ public void setArchiveProcessing(Processing archiveProcessing) {
...
+    /**
+     * Retrieves the archive processing type.
+     * @return The archive processing type.
+     */
+    public Processing getStandardProcessing() {
+        return standardProcessing == null ? Defaults.STANDARD_PROCESSING : standardProcessing;
+    }
+
+    /**
+     * Sets the archive processing type.  If not set will default to NOTIFICATION.
```
ottlinger, line 139: I do not see this default setting in the code or is it handled somewhere else as "if null NOTIFICATION"?

apache-rat-core/src/main/java/org/apache/rat/analysis/DefaultAnalyserFactory.java
```diff
@@ -74,25 +78,48 @@ public DefaultAnalyser(ReportConfiguration config, final Collection<ILicense> li
             this.configuration = config;
         }
 
+        /**
+         * Generates a predicateo to filter out licnses that should not be reported.
```
ottlinger, line 82: LHF: typo predicateo / licnes

apache-rat/src/site/apt/index.apt.vm
```diff
@@ -106,6 +104,7 @@ usage: java -jar apache-rat/target/apache-rat-${project.version}.jar
                                       external xsl file may be specified or one of the internal named sheets: plain-rat (default),
                                       missing-headers, or unapproved-licenses
     --scan-hidden-directories         Scan hidden directories
+    --standard <ProcessingType>       Specifies the level of detail in STANDARD file reporting.. (default is ABSENCE)
```
ottlinger, line 107: LHF: typo - double dot. Should the dot be at the end of the sentence.

src/changes/changes.xml
```diff
@@ -72,6 +72,15 @@ https://maven.apache.org/plugins/maven-changes-plugin/xsd/changes-1.0.0.xsd
     </release>
     -->
     <release version="0.17-SNAPSHOT" date="xxxx-yy-zz" description="Current SNAPSHOT - release to be done">
+      <action issue="RAT-377" type="add" dev="claudenw">
+        Added ability to specify the leverl of reporting on STANDARD files within a project.  This necessitated an addition
```
ottlinger, line 76: LHF: typo leverl ;) Sounds Austrian to my ear.

ottlinger (approved):
Thanks for the change ... apart from my minor things, looks good to me.

apache-rat-core/src/main/java/org/apache/rat/Report.java
Claudenw: We don't actually open the file here.  We create an IOSupplier that we will use to open the stream later.  The IOSupplier gives us the ability to use the `try (closable thing) {` block.

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
Claudenw: `getStandardProcessing` has the check for null returning Defaults.STANDARD_PROCESSING (which is currently set to NOTIFICATION)

apache-rat-core/src/main/java/org/apache/rat/Report.java
ottlinger: ok

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
ottlinger: Must have been to late when I did the review. Thanks.

Claudenw:
I had to rework some code to bring the errors down below the minimum.  The net result is a much larger change.  You may want to review  the last commit.  Please reapprove when you are satisfied.

apache-rat-core/src/main/java/org/apache/rat/license/LicenseSetFactory.java
```diff
@@ -75,7 +77,8 @@ public LicenseSetFactory(SortedSet<ILicense> licenses, Collection<String> approv
      * @return An empty sorted set of ILicense objects.
      */
     public static SortedSet<ILicense> emptyLicenseSet() {
-        return new TreeSet<>(ILicense.getComparator());
+        //return new TreeSet<>((a,b) -> a.getLicenseFamily().compareTo(b.getLicenseFamily()));
```
ottlinger, line 80: LHF: rm old code

apache-rat-core/src/main/java/org/apache/rat/license/LicenseSetFactory.java
```diff
@@ -161,8 +164,9 @@ public SortedSet<String> getLicenseFamilyIds(LicenseFilter filter) {
      * @param licenses the SortedSet of licenses to search.
      * @return the matching license or {@code null} if not found.
      */
-    public static ILicense search(String licenseId, SortedSet<ILicense> licenses) {
-        ILicenseFamily searchFamily = ILicenseFamily.builder().setLicenseFamilyCategory(licenseId)
+    public static Optional<ILicense> search(String familyId, String licenseId, SortedSet<ILicense> licenses) {
+       // return licenses.stream().filter( l -> l.getId().equals(licenseId)).findFirst();
```
ottlinger, line 168: LHF: rm old code?

apache-rat-core/src/main/java/org/apache/rat/license/LicenseSetFactory.java
```diff
@@ -212,13 +221,14 @@ public IHeaderMatcher getMatcher() {
 
     /**
      * Search a SortedSet of licenses for the matching license.
+     * License must mach both family code, and license id.
```
ottlinger, line 224: Typo: mach / match

apache-rat-core/src/test/java/org/apache/rat/analysis/matchers/SPDXMatcherTest.java
```diff
@@ -22,22 +22,32 @@
...
+        target1.reset();
     }
 
     @Test
     public void testMatch() {
-        assertFalse(target.matches(AbstractMatcherTest.makeHeaders("SPDX-License-Identifier: Apache-2", null)));
-        assertTrue(target.matches(AbstractMatcherTest.makeHeaders("SPDX-License-Identifier: hello", null)));
-        target.reset();
+        StringBuilder sb = new StringBuilder()
+                .append("SPDX-License-Identifier: world").append(System.lineSeparator())
```
ottlinger, line 43: constant from above could be used in these tests as well.

apache-rat-plugin/src/test/java/org/apache/rat/mp/RatCheckMojoTest.java
```diff
@@ -200,7 +200,7 @@ public void testIt5() throws Exception {
         assertNotNull(LicenseFamilySetFactory.search("YAL", config.getLicenseFamilies(LicenseFilter.ALL)));
         ReportConfigurationTest.validateDefaultLicenses(config, "MyLicense", "CpyrT", "RegxT", "SpdxT", "TextT", 
                 "Not", "All", "Any");
-        assertNotNull(LicenseSetFactory.search("MyLicense", config.getLicenses(LicenseFilter.ALL)));
+        assertTrue(LicenseSetFactory.search("YAL", "MyLicense", config.getLicenses(LicenseFilter.ALL)).isPresent());
```
ottlinger, line 203: Could we use assertJ's assertThat(...).isPresent here?

ottlinger (approved):
Thanks for taking care of the spotbugs findings, only minor things marked in this review run. Thanks again!

apache-rat-plugin/src/test/java/org/apache/rat/mp/RatCheckMojoTest.java
Claudenw: I don't think that is the case.  Is present() is called because an Optional is returned.  I did a quick change but the result was a line of the same length and complexity.

ottlinger: np, I just thought that 
https://javadoc.io/static/org.assertj/assertj-core/3.24.2/org/assertj/core/api/AbstractOptionalAssert.html#isPresent()
assertThat(LicenseFactory.search(....)).isPresent() would be easier to read. Thanks.

PR #251, opened 2024-05-11
RAT-369: Let the build fail in case of new violations/spotbugs errors

ottlinger:
@Claudenw I was unable to read the docs properly - there is a solution that I`ve added that lets the build fail if new errors are added. At the moment there are 63 open issues within RAT.

Feel free to hit the merge button if that is fine with you.


ottlinger:
Funny catch that newer JDK spot more errors in RAT, thus I had to raise the limit to comply with JDK21.

Claudenw (approved):
Thanks for the cleanup.

PR #264, opened 2024-06-04
WIP RAT-336: Start to migrate tests in first module

ottlinger:
Starting to migrate to AssertJ assertions .....


apache-rat-core/src/test/java/org/apache/rat/analysis/license/AbstractLicenseTest.java
```diff
@@ -110,8 +107,7 @@ public void testEmbeddedStrings(String id, String family, String name, String no
                 for (String fmt : formats) {
                     boolean found = processText(license, String.format(fmt, target[TEXT]));
                     license.reset();
-                    assertTrue(found, () -> String.format("%s %s did not match pattern '%s' for target string %s", id,
```
ottlinger, line 113: @Claudenw it seems that assertJ has no equivalent to this supplier construct.
Do you have an idea how to transform these kind of tests into assertJ or should we leave the other assertTrue() - WDYT?

Claudenw: You could generate the message string and just use it.  It will have some performance impact.  Or you could just leave the junit 5 format

ottlinger: I never worked with that construct before and wondered what it is good for. Thx.

PR #285, opened 2024-07-06
RAT-323, RAT-378 : Harmonize  - Part 3

Claudenw:
Fixes for RAT 378
Part of RAT-323

The major changes here are: 

- The Option definitions have been moved to a new enumeration `commandline/Arg.java`.  In fact the entire `commandline` package is new.  In addition the `Arg` class uses commons/command-cli `OptionGroup` objects to store the options.  This gives us a consistent structure to use without having to check for `Option` vs `OptionGroup` in the `Args` structure.  
- The management of included / excluded licenses is now handled by the `LicenseSetFactory`.  This includes new code to support adding/removing licenses separate from license categories.  Previously filtering was only done by category.  This supports a finer grained approach.  Meaning that we can now list all ASF licenses under ASF and still only approve version 2.0 for use.
- Additional cleanup of the earlier URL -> URI change.
- Ordering of imports, adding "final" to arguments, adding package-info.java files to keep checkstyle happy.
- Switching to always using stylesheets.  XML format is now a stylesheet that only pretty prints.  This simplifies the options around styling and providing style sheets.
- Added `commandline/StyleSheets` to manage access to the stylesheets both internal and externally declared.


ottlinger:
@Claudenw pls merge in master and rebase so that we get all the latest changes from master and part2 properly. Thanks.

apache-rat-tools/src/test/java/org/apache/rat/tools/NamingTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.io.File;
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class NamingTest {
+    @Test
+    public void testNaming() throws IOException {
+        File d = new File("target/com/example");
```
ottlinger, line 31: Personally I would prefer to use TestFolders here so that we do not pollute the target folder with "random" test data.

apache-rat-tools/src/test/java/org/apache/rat/tools/AntGeneratorTest.java
```diff
@@ -0,0 +1,36 @@
...
+import java.io.File;
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class AntGeneratorTest {
+
+    @Test
+    public void testGenerateAntProject() throws IOException {
+        AntGenerator.main( new String[]{"com.example", "AntExample", "target"});
```
ottlinger, line 32: Testfolder?

apache-rat-core/src/test/java/org/apache/rat/commandline/InputArgsTest.java
```diff
@@ -0,0 +1,156 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class InputArgsTest {
+
+    /** The base directory for the test.  We do not use TempFile because we want the evidence of the run to exist after
+     * a failure.*/
+    @TempDir
```
ottlinger, line 53: Is the javadoc wrong or do we need to change to not use TempDir here in case we want to keep the evidence?

apache-rat-tools/src/test/java/org/apache/rat/tools/NamingTest.java
ottlinger: Left unchanged as it is the tools submodule.

apache-rat-tools/src/test/java/org/apache/rat/tools/AntGeneratorTest.java
ottlinger: Left unchanged as it is the tools submodule.

apache-rat-core/src/test/java/org/apache/rat/commandline/InputArgsTest.java
ottlinger: Fixed

PR #299, opened 2024-08-03
RAT-394: fix to make all code use DefaultLog (Harmonize - Part 4)

Claudenw:
Part of RAT-323

Remove most passing of LOG instances in favour of using DefaultLog.getInstance() to retrieve it where necessary.

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
@@ -682,10 +682,12 @@ public enum Arg {
      * @return The Arg or {@code null} if no Arg is found.
      */
     public static Arg findArg(Option optionToFind) {
-        for (Arg arg : Arg.values()) {
-            for (Option candidate : arg.group.getOptions()) {
-                if (optionToFind.equals(candidate)) {
-                    return arg;
+        if (optionToFind != null) {
+            for (Arg arg : Arg.values()) {
+                for (Option candidate : arg.group.getOptions()) {
+                    if (optionToFind.equals(candidate)) {
+                        return arg;
+                    }
                 }
             }
         }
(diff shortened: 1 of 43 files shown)


apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
```diff
@@ -659,4 +675,66 @@ public static void reset() {
...
+        }
+        return null;
+    }
+
+    /**
+     * Finds the Arg that contains an Option with the specified key.
+     * @param key the key for the Option to locate.
+     * @return The Arg or {@code null} if no Arg is found.
+     */
+    public static Arg findArg(String key) {
```
ottlinger, line 700: Is NPE-safeness an issue here?

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
```diff
@@ -659,4 +675,66 @@ public static void reset() {
             }
         }
     }
+
+    /**
+     * Finds the Arg that an Option is in.
+     * @param optionToFind the Option to locate.
+     * @return The Arg or {@code null} if no Arg is found.
+     */
+    public static Arg findArg(Option optionToFind) {
```
ottlinger, line 684: Is NPE-safeness an issue here?

ottlinger:
@Claudenw except from the 2 NPE-questions everything looks fine. I had the impression that we introduced the Log and are now removing it again .....

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
Claudenw: Yes. fixed

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
Claudenw: Yes. fixed

ottlinger:
Thanks for all your work - pls refresh the other pull requests to fetch changes from master.

PR #311, opened 2024-09-05
RAT-390 migrate exclusion config - (Harmonize Part 6)

Claudenw:
Moved exclusion calculations from plugin to core.

Involved refactoring and reimplementing to better fit into the design.

Design now has an ExclusionProcessor that correctly develops a java.nio.PathMatcher which is used to filter files for processing.

Added enhanced iterator library from Jena with minor modifications to suit our needs.
Added and enhanced code from MatchPattern, MatchPatterns,  and SelectorUtils from plexus code base.
Added some rules for checkstyle and for some checkstyle ignore rules.

Added tests.


ottlinger:
The changes seem to have induced test irregularities .... not sure if you spot the cause, @Claudenw ?
https://github.com/Claudenw/creadur-rat/actions/runs/11262029997/job/31316786387

The changes are really great (still reviewing/reading) :)

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -418,7 +444,8 @@ public void setOut(final File file) {
                 DefaultLog.getInstance().warn("Unable to delete file: " + file);
             }
         }
-        if (!file.getParentFile().mkdirs()) {
+        File parent = file.getParentFile();
+        if (!parent.mkdirs() && !parent.isDirectory()) {
```
ottlinger, line 448: When reading this line I stumbled if it can be simplified ....

NOT(a) and NOT(b) == a OR b ?!

But I'm not sure if it is easier to read as 
`!parent.mkdirs() && parent.isFile()` .... WDYT?

Claudenw: I can go either way.  I wrote it as not(isDirectory) because the message also speaks in terms of "directory"

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
```diff
@@ -0,0 +1,306 @@
...
+     * The files and directories created by a CVS source code control based tool.
+     */
+    CVS("The files and directories created by a CVS source code control based tool.",
+            Arrays.asList("**/.cvsignore",
+                    "**/RCS/**", "**/SCCS/**", "**/CVS/**", "**/CVS.adm/**",
+                    "**/RCSLOG/**", "**/cvslog.*", "**/tags/**", "**/TAGS/**",
+                    "**/.make.state", "**/.nse_depinfo",
+                    "**/*~", "**/#*", "**/.#*", "**/,*", "**/_$*", "**/*$", "**/*.old", "**/*.bak", "**/*.BAK",
+                    "**/*.orig", "**/*.rej", "**/.del-*",
+                    "**/*.a", "**/*.old", "**/*.o", "**/*.obj", "**/*.so", "**/*.exe",
```
ottlinger, line 70: Does *.exe really relate to CVS? Where does the list come from?

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
```diff
@@ -0,0 +1,306 @@
...
+            Collections.singletonList("**/SCCS/**"), null, null),
+    /**
+     * The files and directories created by a Serena Dimensions V10 change control system based tool.
+     */
+    SERENA_DIMENSIONS_10("The files and directories created by a Serena Dimensions V10 change control system based tool.",
+            Collections.singletonList("**/.metadata/**"), null, null),
+    /**
+     * A standard collection of generally accepted patterns to ignore.
+     */
+    STANDARD_PATTERNS("A standard collection of generally accepted patterns to ignore.", null, null, null),
```
ottlinger, line 173: Should we document that both standard-things are empty?! or add a comment why they are necessary? By reading over it I asked myself what they are good for.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
```diff
@@ -0,0 +1,306 @@
...
+    /**
+     * A standard collection of SCMs
+     */
+    STANDARD_SCMS("A standard collection of SCMs", null, null, null),
```
ottlinger, line 177: Maybe the method ` private Set<StandardCollection> getCollections()` should be referenced here as it calculates the standard collections and motivates why the enum's lists are all null?!

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/DescendingFileProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+ * A FileProcessor that assumes the files contains the already formatted strings and just need to be
+ * localized for the fileName.
+ */
+public class DescendingFileProcessor implements FileProcessor {
+    /** The name of the file being processed */
+    private final String fileName;
+    /** the predicate that will return {@code false} for any comment line in the file. */
+    protected final Predicate<String> commentFilter;
+    /** The system dependant file name separator.  e.g. "/" on linux. */
+    protected final String separator;
```
ottlinger, line 46: Should we use JDK-constants here instead of reinventing the wheel? e.g. https://docs.oracle.com/javase/8/docs/api/java/io/File.html's separator constants

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/GitFileProcessor.java
```diff
@@ -0,0 +1,53 @@
...
+ * under the License.
+ */
+package org.apache.rat.config.exclusion.fileProcessors;
+
+import org.apache.rat.document.impl.DocumentName;
+
+/**
+ * @see <a href='https://git-scm.com/docs/gitignore'>.gitignore documentation</a>
+ */
+public class GitFileProcessor extends DescendingFileProcessor {
```
ottlinger, line 26: Will the library we have introduced in release 0.16.0 onwards be used to perform the gitignore parsing or can this lib be removed completely by your PR?

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/plexus/SelectorUtils.java
```diff
@@ -0,0 +1,720 @@
...
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+ * OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ * ====================================================================
+ *
+ * This software consists of voluntary contributions made by many
+ * individuals on behalf of the Apache Software Foundation.  For more
+ * information on the Apache Software Foundation, please see
+ * <http://www.codehaus.org/>.
```
ottlinger, line 54: Sounds contradictory to me as codehaus.org is not a ASF-site. Is it a problem that we include a ASF1.1-file in our ASF2.0-codebase?

apache-rat-core/src/main/java/org/apache/rat/document/impl/DocumentName.java
```diff
@@ -0,0 +1,220 @@
...
+    private final String baseName;
+    /** The directory separator for this document. */
+    private final String dirSeparator;
+    /** The case-sensitive flag */
+    private final boolean isCaseSensitive;
+
+    static {
+        boolean fsSensitive = true;
+        try {
+            Path p = Files.createTempDirectory("NameSet");
```
ottlinger, line 53: Is this safe on multiple runs or should we add more random elements to the tmp-dir name here?

ottlinger: Apart from that I wondered how to make sure that the NameSet-dir is really removed in case of errors or start/shutdown scenarios ....

apache-rat-core/src/main/java/org/apache/rat/document/impl/DocumentName.java
```diff
@@ -0,0 +1,220 @@
...
+        parts.addAll(Arrays.asList(tokenize(child)));
+        String newName = String.join(dirSeparator, parts);
+        return new DocumentName(newName, baseName, dirSeparator, isCaseSensitive);
+    }
+
+    /**
+     * Gets the fully qualified name of the document.
+     * @return the fully qualified name of the document.
+     */
+    public String name() {
```
ottlinger, line 111: to be compliant with the Java-rules this should be getName() as the class is not an enum - WDYT?

apache-rat-core/src/main/java/org/apache/rat/document/impl/DocumentName.java
```diff
@@ -0,0 +1,220 @@
...
+        return name;
+    }
+
+    /**
+     * Gets the fully qualified basename of the document.
+     * @return the fully qualified basename of the document.
+     */
+    public String baseName() {
```
ottlinger, line 119: to be compliant with the Java-rules this should be getBaseName() as the class is not an enum - WDYT?

apache-rat-core/src/main/java/org/apache/rat/document/impl/DocumentName.java
```diff
@@ -0,0 +1,220 @@
...
+        return baseName;
+    }
+
+    /**
+     * Gets the DocumentName for the basename of this document name.
+     * @return the DocumentName for the basename of this document name.
+     */
+    public DocumentName baseDocumentName() {
```
ottlinger, line 127: to be compliant with the Java-rules this should be getDocumentName() as the class is not an enum - WDYT?

apache-rat-core/src/main/java/org/apache/rat/document/impl/DocumentName.java
```diff
@@ -0,0 +1,220 @@
...
+        return name.equals(baseName) ? this : new DocumentName(baseName, baseName, dirSeparator, isCaseSensitive);
+    }
+
+    /**
+     * Returns the directory separator.
+     * @return the directory separator.
+     */
+    public String dirSeparator() {
```
ottlinger, line 135: to be compliant with the Java-rules this should be getDirSeparator() as the class is not an enum - WDYT?

ottlinger: Applies to some more methods in this class .....

apache-rat-plugin/pom.xml
```diff
@@ -415,7 +415,7 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
-      <version>2.0.16</version>
+      <version>2.0.13</version>
```
ottlinger, line 418: Why is this downgrade necessary or is it a merge issue in this branch?

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/GitFileProcessor.java
ottlinger: I mean 
```
    <dependency>
      <groupId>nl.basjes.gitignore</groupId>
      <artifactId>gitignore-reader</artifactId>
      <version>1.6.0</version>
    </dependency>
```

Claudenw: I believe it can be removed.  The new mechanism does not build a list of all files at start and just does filtering as the file is discovered so it should be faster startup.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
Claudenw: https://www.gnu.org/software/trans-coord/manual/cvs/html_node/cvsignore.html#cvsignore

Claudenw: added info in javadoc

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
Claudenw: added java comment -- not javadoc.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/DescendingFileProcessor.java
Claudenw: Good catch.  I think that it is actually a leftover from an earlier implementation and is no longer used.

apache-rat-core/src/main/java/org/apache/rat/document/impl/DocumentName.java
Claudenw: good catch, A added a finally block to the try.

apache-rat-core/src/main/java/org/apache/rat/document/impl/DocumentName.java
Claudenw: ok

apache-rat-core/src/main/java/org/apache/rat/document/impl/DocumentName.java
Claudenw: ok

apache-rat-plugin/pom.xml
Claudenw: merge issue

apache-rat-core/src/main/java/org/apache/rat/document/impl/DocumentName.java
Claudenw: fixed

apache-rat-core/src/main/java/org/apache/rat/document/impl/guesser/NoteGuesser.java
```diff
@@ -45,26 +52,37 @@ public class NoteGuesser {
         "THIRD_PARTY_NOTICES", "THIRD_PARTY_NOTICES.TXT",
         "COPYRIGHT", "COPYRIGHT.TXT",
         "BUILDING", "BUILDING.TXT",
-        "BUILD", "BUILT.TXT",//
+        "BUILD", "BUILT.TXT",
```
ottlinger, line 55: Does it make sense to add markdown as extension as well? md/MD?

apache-rat-core/src/main/java/org/apache/rat/document/impl/guesser/NoteGuesser.java
```diff
@@ -45,26 +52,37 @@ public class NoteGuesser {
...
         "DEPENDENCIES"
     };
 
+    /**
+     * List of note file extensions.  Extensions that indicate a file is a note file.
+     */
     private static final String[] NOTE_FILE_EXTENSIONS = {
```
ottlinger, line 62: Just by following the name I'd guess that extensions should mean:
* empty/none
* .txt
* .md
and we could iterate over file names and its extensions - WDYT?

apache-rat-core/src/main/java/org/apache/rat/utils/iterator/FilterIterator.java
```diff
@@ -0,0 +1,97 @@
...
+package org.apache.rat.utils.iterator;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+/**
+ Creates a sub-Iterator by filtering. This class should not be used
+ directly any more; the subclasses FilterKeepIterator and FilterDropIterator
```
ottlinger, line 28: Do these classes exist or is the comment outdated?

apache-rat-core/src/main/java/org/apache/rat/utils/iterator/NiceIterator.java
```diff
@@ -0,0 +1,234 @@
...
+
+
+/**
+ NiceIterator is the standard base class implementing ExtendedIterator. It provides
+ the static methods for <code>andThen</code>, <code>filterKeep</code> and
+ <code>filterDrop</code>; these can be reused from any other class. It defines
+ equivalent instance methods for descendants and to satisfy ExtendedIterator.
+ */
+
+public class NiceIterator<T> implements ExtendedIterator<T> {
```
ottlinger, line 40: Shouldn't this be an EmptyIterator instead?

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/GitFileProcessor.java
ottlinger: Dpeendency removed. RAT still compiles ;)

apache-rat-core/src/main/java/org/apache/rat/document/impl/guesser/NoteGuesser.java
Claudenw: Yes it would.  But the NoteGuesser needs to be rethought.  It should work like the license matchers do.  For example we want to be able to say that license X requires a copyright notice.  So read the notice files and see if the copyright is found.

I think we have the components to make this work much better but it is not part of this PR/ticket.

apache-rat-core/src/test/java/org/apache/rat/document/impl/guesser/NoteGuesserTest.java
```diff
@@ -34,34 +39,24 @@ public void testMatches() {
...
-        assertTrue(NoteGuesser.isNote("src/test/README.txt"));
-        assertTrue(NoteGuesser.isNote("src\\test\\DEPENDENCIES"));
-        assertTrue(NoteGuesser.isNote("src\\test\\LICENSE"));
-        assertTrue(NoteGuesser.isNote("src\\test\\LICENSE.txt"));
-        assertTrue(NoteGuesser.isNote("src\\test\\NOTICE"));
-        assertTrue(NoteGuesser.isNote("src\\test\\NOTICE.txt"));
-        assertTrue(NoteGuesser.isNote("src\\test\\README"));
-        assertTrue(NoteGuesser.isNote("src\\test\\README.txt"));
+
+    @Disabled("RAT-390: old tests that do not work anymore")
```
ottlinger, line 52: @Claudenw shouldn't these cases work as the normalisation is supposed to remove path separators? I just reformatted the commented-out stuff and added it here as the tests with backslashes do not recognise the files as a note.

Claudenw: Short answer -- no. 

Long answer.  The TestingDocument assumes that it is running on a linux box if you don't tell it otherwise.  To make the test work on any platform we need to create a DocumentName with the proper separator character..  The new I modified TestingDocument to accept a DocumentName and modified the NoteGuesserTest to generate document names in Linux and Windows formats.  The base name is the point from which the search for files started.  In the test case I used the root directory though the "src" directory could have been used for the ones with paths.

Once the DocumentNames are set correctly it works.

apache-rat-core/src/main/java/org/apache/rat/document/impl/guesser/NoteGuesser.java
ottlinger: @Claudenw maybe we leave this for a different kind of refactoring, but I still find the name a bit misleading :) (but will merge nevertheless) Thx.

ottlinger:
@Claudenw kudos for this great work! Let us see how downstream users benefit from this new and enhanced functionality.

PR #339, opened 2024-10-04
RAT-430: fixed checkstyle issues in report/xml

Claudenw:
fix for RAT-430
Ignored magic numbers nad javadoc in old XmlWriter code

Leaves an issue with `report/xml/writer/impl/base` not having a `package-info.java` file, but I think that we should probably move all the files into `report/xml/writer` after RAT-323 is completed


ottlinger (approved):
LGTM

PR #360, opened 2024-10-22
RAT-419: Checkstyle fixes for config/parameters package

Claudenw:
Fix for RAT-419

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
@@ -61,15 +61,6 @@ public class Description {
     private final Map<String, Description> children;
 
     /**
-     * A predicate to filter Descriptions by component type.
-     * @param type The component type required in the filter.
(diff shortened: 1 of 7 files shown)


apache-rat-core/src/main/java/org/apache/rat/config/parameters/Description.java
```diff
@@ -106,7 +106,7 @@ public Description(final ComponentType type, final String name, final String des
 
     /**
      * Constructor
-     * @param configComponent the configuration component
+     * @param configComponent the configuration component.
```
ottlinger, line 109: @Claudenw class has unused methods - remove tpePredicate()?! Thx.

Claudenw: I think this got cleaned up in RAT-420.  If you merge that one first it might get rid of this.

ottlinger: I manually removed this method.

ottlinger (approved):
LGTM

PR #361, opened 2024-10-22
RAT-420: fixed checkstyle and removed unused methods in config package

Claudenw:
Fix for RAT-420


apache-rat-core/src/main/java/org/apache/rat/config/exclusion/ExclusionUtils.java
```diff
@@ -280,7 +280,7 @@ public static Optional<MatchPatterns> notMatchPattern(final Iterator<String> ite
 
     /**
      * @param filename file name to cleanup.
-     * @return
+     * @return cleaned up file name.
```
ottlinger, line 283: @Claudenw the class has multiple unused methods - should we just remove them?
(same in other classes in this package/subpackage)

Claudenw: Removing methods always makes me nervous.  But I am going to say yes, remove them.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/ExclusionUtils.java
```diff
@@ -209,110 +185,4 @@ private static void verifyFile(final File file) {
             throw new ConfigurationException(format("%s is not a valid file.", file));
         }
     }
-
```
ottlinger, line 212: @Claudenw I'm a bit suspicious if I removed too much ....
some of the methods are used by Reflection in other PRs and failed the build. But the build runs through .... is that code really unused or am I missing something here?

apache-rat-core/src/main/java/org/apache/rat/config/AddLicenseHeaders.java
```diff
@@ -23,25 +23,10 @@
...
-    FALSE, 
-    /** Add headers and overwrite the old file */
-    FORCED;
-
-    private static final char SEPARATOR = '|';
-
-    /**
-     * @return pretty-printed list of all available options.
-     */
-    public static String getValuesForHelp() {
```
ottlinger, line 38: @Claudenw does a call to this method miss somewhere in the utils module in order to print all available methods?

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/ExclusionUtils.java
Claudenw: I think it is fine.

PR #366, opened 2024-10-27
RAT-422:  checkstyle fixes for configuration

Claudenw:
Also includes reinstatement of method called by reflection.

Fix for RAT-422


apache-rat-core/src/main/java/org/apache/rat/configuration/XMLConfig.java
```diff
@@ -23,36 +23,36 @@
...
+    public static final String LICENSE = "license";
     /** approved element name */
-    public final static String APPROVED = "approved";
+    public static final String APPROVED = "approved";
     /** family element name */
-    public final static String FAMILY = "family";
+    public static final String FAMILY = "family";
     /** note element name */
-    public final static String NOTE = "note";
+    public static final String NOTE = "note";
```
ottlinger, line 51: This constant is unused - does it get called via reflection or can we safely remove it here? @Claudenw

ottlinger (approved):
LGTM, only thsi one question concerning the NOTE constant.

PR #367, opened 2024-10-27
RAT-423: Fixed checkstyle and moved DocumentAnalyserMultiplexer

Claudenw:
Fixes for RAT-423


apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -41,9 +41,9 @@
 import org.apache.rat.config.AddLicenseHeaders;
 import org.apache.rat.config.exclusion.ExclusionProcessor;
 import org.apache.rat.config.exclusion.StandardCollection;
-import org.apache.rat.document.impl.DocumentName;
-import org.apache.rat.document.impl.DocumentNameMatcher;
-import org.apache.rat.document.impl.DocumentNameMatcherSupplier;
+import org.apache.rat.document.DocumentName;
```
ottlinger, line 44: @Claudenw Class has unused methods addIncludedFilter/addFamilies/addApprovedLicenseId() - do we need a reflection test for these as well or are they really unused?

ottlinger (approved):
LGTM, only this one unused code question remains

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
Claudenw: These are convenience methods.  I will add tests.

PR #375, opened 2024-10-28
RAT-435: checkstyle fixes for root package

Claudenw:
fix for RAT-435


apache-rat-core/src/main/java/org/apache/rat/BuilderParams.java
```diff
@@ -33,7 +33,7 @@ public interface BuilderParams {
 
     /**
      * Gets one of the contained methods by name.
-     * 
+     *
```
ottlinger, line 36: matcherMap() and licenseFamilies() are reported as never used - is there a test case missing or is it called by reflection as well, @Claudenw ?

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -279,15 +279,15 @@ public void addExcludedCollection(StandardCollection collection) {
      * @param collection the StandardCollection to exclude.
      * @see ExclusionProcessor#addFileProcessor(StandardCollection)
      */
-    public void addExcludedFileProcessor(StandardCollection collection) {
+    public void addExcludedFileProcessor(final StandardCollection collection) {
```
ottlinger, line 282: What to do with unused methods in here ..... (possibly referenced in other PRs) @Claudenw

ottlinger (approved):
Apart from the unused methods in ReportConfiguration looks ok. Thanks

apache-rat-core/src/main/java/org/apache/rat/BuilderParams.java
Claudenw: I added comments to indicate the methods are called by reflection and also updated PR #366 (RAT-422) to test that the methods in the annotation exist in the BuilderParams class.

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
Claudenw: Handled in RAT-423

PR #376, opened 2024-10-28
RAT-438: Cleaned up checkstyle and associated issues

Claudenw:
Fixes for RAT-438


apache-rat-tools/src/main/java/org/apache/rat/tools/AntDocumentation.java
```diff
@@ -55,8 +47,6 @@
  * A simple tool to convert CLI options into an Ant report base class.
  */
 public final class AntDocumentation {
```
ottlinger, line 49: @Claudenw class contains 3 unused methods - are these called by reflection as well and we need some more tests?

apache-rat-tools/src/main/java/org/apache/rat/tools/MavenGenerator.java
```diff
@@ -171,7 +171,7 @@ private static String getComment(final MavenOption option) {
                 throw new IllegalStateException(format("Argument type %s must be in OptionCollection.ARGUMENT_TYPES", option.getArgName()));
             }
             String typeDesc = sup.get();
-            typeDesc = WordUtils.uncapitalize(typeDesc.substring(0,1)) + typeDesc.substring(1);
+            typeDesc = WordUtils.uncapitalize(typeDesc.substring(0, 1)) + typeDesc.substring(1);
```
ottlinger, line 174: this value is not used in the code that follows this line? What does it do ?

apache-rat-tools/src/main/java/org/apache/rat/tools/xsd/XsdGenerator.java
```diff
@@ -73,10 +68,8 @@ public class XsdGenerator {
      * @throws IOException on IO errors.
      * @throws TransformerException if the XSD can not be pretty printed.
      */
-    public static void main(String[] args) throws IOException, TransformerException {
-        ReportConfiguration configuration = OptionCollection.parseCommands(args, (options) -> {
-        });
-        XsdGenerator generator = new XsdGenerator(configuration);
+    public static void main(final String[] args) throws IOException, TransformerException {
```
ottlinger, line 71: @Claudenw is it correct to remove the configuration here? It seems to be unused.

apache-rat-tools/src/main/java/org/apache/rat/tools/MavenGenerator.java
Claudenw: Good catch.  It was a holdover from earlier when the type info was displayed inline in the description rather than in a table below.

apache-rat-tools/src/main/java/org/apache/rat/tools/xsd/XsdGenerator.java
Claudenw: Yes, it was a holdover from an earlier process.

ottlinger (approved):
Aparet from the unused stuff, LGTM - @Claudenw

ottlinger:
I do not know if it is a bug or a feature, but if I click on certain menu elements (=i.e. documentation pages) its menu item is not properly highlighted - I assume this to be a situation irrelevant to the changes and will have a look at it, when updating to the new Doxia 2.x/Maven4 site stack.

PR #380, opened 2024-11-01
RAT-438: checkstyle fixes for apache-rat-tasks

Claudenw:
Removed deprecated code from consideration


ottlinger (approved):
Is this all here? @Claudenw

Claudenw:
Yep.  It was just a case of fixing the exclude statement

PR #381, opened 2024-11-01
RAT-436: checkstyle fixes for plugin

Claudenw:
Fix for RAT-436
Fixed checkstyle issues
    

- Marked deprecated code
- Removed deprecated code from checkstyle consideration
- Fixed checkstyle issues.
- Fixed javadoc issue.

### DIFF
--- a/apache-rat-plugin/src/main/java/org/apache/rat/mp/HeaderMatcherSpecification.java
+++ b/apache-rat-plugin/src/main/java/org/apache/rat/mp/HeaderMatcherSpecification.java
@@ -22,7 +22,9 @@ package org.apache.rat.mp;
 /**
  * Used to specify additional license matchers. A license matcher is basically a class, which implements
  * {@link org.apache.rat.analysis.IHeaderMatcher}.
+ * @deprecated use Configuration file to define licenses
  */
+@Deprecated
 public class HeaderMatcherSpecification
 {
     /**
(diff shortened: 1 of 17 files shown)


apache-rat-plugin/src/main/java/org/apache/rat/mp/HeaderMatcherSpecification.java
```diff
@@ -22,7 +22,9 @@
 /**
  * Used to specify additional license matchers. A license matcher is basically a class, which implements
  * {@link org.apache.rat.analysis.IHeaderMatcher}.
+ * @deprecated use Configuration file to define licenses
  */
+@Deprecated
 public class HeaderMatcherSpecification
```
ottlinger, line 28: This class contains unused methods? Is it reflection as well @Claudenw ?

ottlinger: Other methods in these classes are marked as unused as well - I assume it's all reflection and means missing tests ....

ottlinger (approved):
I assume the unused stuff is not so relevant here as the code is deprecated.

apache-rat-plugin/src/main/java/org/apache/rat/mp/HeaderMatcherSpecification.java
Claudenw: Yes, these were called by entries in the pom.xml file.

PR #387, opened 2024-11-09
RAT-399: migrate ignore errors

Claudenw:
Fix for RAT-399 and RAT-396

Moved the ignore error code into core and applied switches to all of the ClaimStatistic.Counter types.

On the command line 
`--counter-max` and `--counter-min` have been added.  They can take multiple arguments.  The arguments have a structure of "<Counter>:<count>".  For example  `--counter-max UNAPPROVED:5" will set the maximum number of allowed UNAPPROVED liceses to 5.

XML output has been modified to report issues with statistics
XSD has been update to reflect the changes.
XSL stylesheets have been updated.

Help has been updated to display the Counter type information.


apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
```diff
@@ -19,34 +19,94 @@
...
+        }
+
+        public String displayName() {
+            return StringUtils.capitalize(name().replaceAll("_", " ").toLowerCase(Locale.ROOT));
+        }
+    }
 
     /** Count of license family name to counter */
-    private final ConcurrentHashMap<String, IntCounter> licenseFamilyNameMap = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, IntCounter> licenseNameMap = new ConcurrentHashMap<>();
```
ottlinger, line 105: Should we get rid of the IntCounter here and use AtomicInteger instead?

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
```diff
@@ -19,46 +19,55 @@
...
 <xsl:stylesheet version="1.0"
                 xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+    <xsl:variable name='newline'><xsl:text>&#xa;</xsl:text></xsl:variable>
+
 <xsl:output method='text'/>
 <xsl:template match='/'>
 *****************************************************
 Summary
--------
+*****************************************************
```
ottlinger, line 28: I assume the removed line was correct here as the asterisks indicate a start of an element? WDYT?

The report starts with asterisks and ends with asterisks.

apache-rat/src/site/examples/default_output.txt
```diff
@@ -1,81 +1,96 @@
...
+Approved:           9    A count of approved licenses.
+Archives:           1    A count of archive files.
+Binaries:           2    A count of binary files.
+Document types:     5    A count of distinct document types.
+Generated:          1    A count of generated files.
+License categories: 5    A count of distinct license categories.
+License names:      6    A count of distinct license names.
+Notices:            2    A count of notice files.
+Standards:          8    A count of standard files.
+Unapproved:         2 (Exceeded limits)    A count of unapproved licenses.
```
ottlinger, line 17: Personally I find this irritating as it breaks the readability of the report. Could we just remove the '(exceeded limits)' completely or introduce a new paragraph with 'Exceeded counters' instead? This would allow an easier parsing of the report instead of mixing valid and invalid counters. 

WDYT?

ottlinger:
@Claudenw does this change also address https://issues.apache.org/jira/browse/RAT-396 ?

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
Claudenw: I explored this possibility.  The map is already a `ConcurrentHashMap` and the calculations take place inside a `compute()`.  The `IntCounter` is a slim wrapper around an `int` making it updatable Integer.  This save the boxing/unboxing that would otherwise occur.

Changing to an AtomicInteger involves more overhead and code to do setting but does not gain anthing in terms of thread safety.

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
Claudenw: I think I misunderstood the visual structure of the output.  This is not my speciality. 
I have opened RAT-453 to discuss and fix the formatting for this output.

Claudenw:
@ottlinger I think it does and have added it to the list of tickets resolved with this request, however it may actually be in RAT-453.

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
ottlinger: ok

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
ottlinger: I just had a look at the whole report and all sections started with asterisk, whiole the ---- was just a headline.

PR #390, opened 2024-11-11
RAT-246: .gitignore in parent dir not honored

Claudenw:
RAT-246 was fixed by an earlier fix.

This change adds an integration test framework for core, javadoc based documentation for same, and a test to show that RAT-246 is no longer an issue.

This change is built upon RAT-399 and can not be merged until after that fix is.

Added additional Debug logging to assist in problem determination.


apache-rat-core/src/it/java/org/apache/rat/ReportTest.java
```diff
@@ -0,0 +1,205 @@
...
+ */
+public class ReportTest {
+
+    private String[] asArgs(final List<String> argsList) {
+        return argsList.toArray(new String[argsList.size()]);
+    }
+
+    @ParameterizedTest(name = "{index} {0}")
+    @MethodSource("args")
+    public void integrationTest(String testName, Document commandLineDoc) throws Exception {
```
ottlinger, line 90: Shouldn't all this parsing functionality be put into the tools-submodule?

ottlinger: I guess this testing infra can be used/or extended in the long term to be useful for performance tests as well.

ottlinger:
Apart from the architectural questions I love this new functionality! Thanks.

ottlinger (approved):
LGTM - thx.

apache-rat-core/src/it/java/org/apache/rat/ReportTest.java
Claudenw: Keep in mind that tools use core.  Tools are tools to help extend core.  Also, the parsing here is specific to how I laid out the testing.   Basically a way to get command line arguments as an array and where to capture and store log and output if not otherwise defined.  And then finally using the core framework to find all the "commandLine.txt" files in the directory tree and assume that they are in the tests we want to run.

PR #393, opened 2024-11-15
RAT-453: quick fix for help display of descriptions for counters

Claudenw:


ottlinger:
@Claudenw anything else to be added here? Do we need a test for the behaviour or is it connect to a different branch? Build is fixed.

Claudenw:
This is just a change to the format of a help message.  There are no tests for the message content, I don't think there need to be.

PR #394, opened 2024-11-16
RAT-408: verify not an issue

Claudenw:
closes RAT-408
test to verify RAT-408 is not an issue


ottlinger:
This is great - thanks for adding it and bringing order into all these old Jira tickets #bigKudos!

PR #395, opened 2024-11-16
RAT-81: Fixed encoding issue causing text files to not be read properly

Claudenw:
Fix for RAT-81 and RAT-96

Fixed encoding issue where text files not in UTF-8 encoding would not be read properly.

Change adds charset to the metadata when it can be discovered.  If not UTF8 is returned.

Added integration test RAT-81 to show reading of UTF8 and IBM037 encoding works.


Claudenw:
Should we add the encoding to the XML output?

apache-rat-core/src/main/java/org/apache/rat/analysis/TikaProcessor.java
```diff
@@ -42,6 +47,7 @@ public final class TikaProcessor {
      * "text" types are already handled somewhere else
      * BINARY unless listed here*/
     private static final Map<String, Document.Type> DOCUMENT_TYPE_MAP;
+    //private static final DefaultDetector DEFAULT_DETECTOR = new DefaultDetector();
```
ottlinger, line 50: Can we remove the default_detector here?

apache-rat-core/src/main/java/org/apache/rat/analysis/TikaProcessor.java
```diff
@@ -133,6 +148,31 @@ public static String process(final Document document) throws RatDocumentAnalysis
...
+     */
+    private static Charset detectCharset(final InputStream stream) throws IOException {
+        CharsetDetector encodingDetector = new CharsetDetector();
+        encodingDetector.setText(stream);
+        CharsetMatch charsetMatch = encodingDetector.detect();
+        if (charsetMatch != null) {
+            try {
+                return Charset.forName(charsetMatch.getName());
+            } catch (UnsupportedCharsetException e) {
+                // do nothing
```
ottlinger, line 165: Should we at least log the exception here in order to debug "unexpected behaviour" when browsing contents?

Claudenw: Good catch.

apache-rat-core/src/main/java/org/apache/rat/analysis/TikaProcessor.java
Claudenw: Sorry, yes. done.

ottlinger:
> Should we add the encoding to the XML output?

Do you mean an encoding entry per file or a general encoding information?

Claudenw:
I added "encoding" if the document is a STANDARD type (e.g. not ARCHIVE,  BINARY, GENERATED, NOTICE, or UNKNOWN)
I added mediaType as well since I was there and so was it.

Now at least people can generate reports for non-standard encodings, or unexpected mediaTypes.

ottlinger (approved):
LGTM - thanks for adding two more attributes and a proper encoding handling to RAT!

If you are okay with my changes, feel free to merge this PR

PR #396, opened 2024-11-17
RAT-14: Verification for RAT-14

Claudenw:
Fix for RAT-14

Added integration test
Modified Reporter to allows exec() call and return a `ClaimStatistic` object.
Added groovy usage of ClaimStatistic to verify results whithout generating output.


ottlinger (approved):
Thanks for the fixes - I reduced and renamed the execute() method as it only seems to be called from within tests and is "internal" from the view of Reporter.
If you are ok with it, go ahead with the merge.

Claudenw:
execute() is the solution for the request to get numbers only from the run.  It needs to be public.  I will adjust this.

ottlinger:
> execute() is the solution for the request to get numbers only from the run. It needs to be public. I will adjust this.

Should we add more tests to indicate this? I was astonished to only find references in the tests, thus I changed the visibility.

Claudenw:
I added tests for `execute()`

ottlinger:
@Claudenw I removed the deletion of test directories and corrected the test assertions for the build to pass. Hope this is okay with you.

PR #397, opened 2024-11-17
RAT-259: implement and verify solution in jira ticket

Claudenw:
Fix for RAT-259
Implements "--input-source", refactored how the configuration stores the files/walkers to process.

Should be cleaner to implement other walkers now.


apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -159,6 +169,55 @@ public ReportConfiguration() {
...
+    }
+
+    /**
+     * Adds a file as a source of files to scan.
+     * The file must be a text file that lists files to be included.
+     * File within the file must be in linux format with a
+     * "/" file separator.
+     * @param file the file to process.
+     */
+    public void addSource(final File file) {
```
ottlinger, line 183: Should we add a check here that ensures no backslashes are provided as parameter?

apache-rat-core/src/main/java/org/apache/rat/document/DocumentNameMatcher.java
```diff
@@ -20,17 +20,206 @@
...
+
+import static java.lang.String.format;
+
+/**
+ * Matches document names.
+ */
+public final class DocumentNameMatcher {
+
+    /** The environment variable to set to enable tracing for the PatternMatcher execution */
+    public static final String ENV_VAR = DocumentNameMatcher.class.getName();
```
ottlinger, line 40: Shouldn't there be a reference to this ENV_VAR if it is used to indicate or change something?

apache-rat-core/src/test/java/org/apache/rat/config/exclusion/FileProcessorTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.fail;
+
+public class FileProcessorTest {
+
+    FileProcessor underTest = new FileProcessor(){
```
ottlinger, line 34: object is never used?

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
Claudenw: The parameter may have backslashes if it is on a Windows box.  But the contents for the file identified by the parameter should use "/" as the separator.  Failure to do so may or may not work depending on a lot of other conditions.   I don't think we can/should check for the presence of the backslash as it might be a valid character in a file name.  OK that is a stretch, but I think we should leave it.

ottlinger: ok, I just thought of it, when I read the Javadocs.

ottlinger (approved):
Looks fine - feel free to merge if you are okay with my changes

PR #398, opened 2024-11-19
RAT-2: Added --input-exclude-size option

Claudenw:
Fix for RAT-2

Added --input-exclude-size <bytes> as an option to skip the scanning of very small files.


ottlinger (approved):
LGTM - if you are okay with my changes go ahead and merge.
I realized that ReportConfiguration contains 2 unused methods and wondered if tests should/could be added for them.

Thanks again

Claudenw:
Opened ticket RAT-459 to address missing tests and other minor problems in ReportConfiguration.

PR #399, opened 2024-11-19
Quick fix for log breaking build

Claudenw:
Added ability to set log level in all cases or lot that it is not supported


ottlinger:
Is that related to RAT-458?

Claudenw:
It is related to RAT-259 and RAT-14.  The issue was that the `ReporterTest` integration test switched out the log file but did not switch it back.  And being a static the change impacted other tests.  

The other problem is that the integration test is not really an integration test.  I should probably move it to `apache-rat` from `apache-rat-core` and make it run after the packaging.  This would make the most sense.  

There is some work to be done to clean up this area of testing.  The tests work but as they are the "integration test" label is a bit of a misnomer.

PR #401, opened 2024-11-20
RAT-107: fix excludes working recursively in submodules and add tests

Claudenw:
Fix for RAT-107

Fixes excludes in standard sets.
Added Maven unit and integration tests.


ottlinger:
Shall I merge or do you want to proceed @Claudenw ?

PR #404, opened 2024-11-27
RAT-455: no GPL licenses allowed in default config

Claudenw:
Fix for RAT-455

Removed GPL licenses from approved license family listing.
Updated tests
Added integration test.


ottlinger:
@Claudenw so if my project uses GPL I need to configure it into my pom to allow it, right?

ottlinger (approved):
LGTM - feel free to merge if you are okay with my changes.

PR #405, opened 2024-11-28
RAT-453: updated XSLT and example output

Claudenw:
Fixes for RAT-453, RAT-464, and RAT-466

Updated XSLT and example output from the default transform.
Added media type and encoding to output


ottlinger:
@Claudenw do you have the build failures locally as well?

Claudenw:
@ottlinger tests are fixed and complete now.

Claudenw:
@ottlinger please take a look at this PR.

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
```diff
@@ -19,59 +19,94 @@
...
+        </xsl:call-template>
+
+        <xsl:text>    (Entries starting with '!' exceed the minimum or maximum values)</xsl:text>
+        <xsl:value-of select='$newline'/>
+        <xsl:for-each select='descendant::statistic'>
+            <xsl:call-template name="statistic">
+                <xsl:with-param name="name"><xsl:value-of select="@name"/></xsl:with-param>
+                <xsl:with-param name="count"><xsl:value-of select="@count"/></xsl:with-param>
+                <xsl:with-param name="description"><xsl:value-of select="@description"/></xsl:with-param>
+                <xsl:with-param name="leadin"><xsl:choose>
```
ottlinger, line 51: What does leadIn mean here?

ottlinger: or does it mean "leading"?

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
```diff
@@ -19,59 +19,94 @@
...
+            <xsl:for-each select='descendant::resource[@type="ARCHIVE"]'>
+                <xsl:value-of select='concat($newline, " ", @name)'/>
+            </xsl:for-each>
+        </xsl:if>
+        <xsl:value-of select="concat($newline, $newline)" />
+        <xsl:call-template name="section">
+            <xsl:with-param name="title">Detail</xsl:with-param>
+        </xsl:call-template>
+        <xsl:text>
+  The line following documents with unapproved licenses will start with a '!'
```
ottlinger, line 109: This line makes me stumble in the report. Can it be simplified to:
* "Lines starting with '!' indicate that the document contains an unapproved license" OR
* "Documents found to contain unapproved licenses are marked as '!'"

If I have a look at the reports being generated during the build of the feature branch, I interpret the '!' to mean "this document contains an unapproved license".

WDYT?

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
```diff
@@ -83,18 +118,74 @@ Archives:
...
- </xsl:for-each>
-*****************************************************
-</xsl:template>
+        <xsl:for-each select='descendant::resource'>
+            <xsl:call-template name="document">
+                <xsl:with-param name="name"><xsl:value-of select='@name'/></xsl:with-param>
+                <xsl:with-param name="mediaType"><xsl:value-of select='@mediaType'/></xsl:with-param>
+                <xsl:with-param name="encoding"><xsl:value-of select='@encoding'/></xsl:with-param>
+                <xsl:with-param name="type"><xsl:value-of select="substring(@type, 1, 1)"/></xsl:with-param>
+                <xsl:with-param name="leadin"><xsl:choose>
```
ottlinger, line 127: same as above? Does it mean leading? If so - should it be called 'prefix' instead?

ottlinger (changes requested):
Apart from the two questions - looks good :)

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
Claudenw: The proposed output looks like this:

```
The line following documents with unapproved licenses will start with a '!'
  The next character identifies the document type.

   char         type
    A       Archive file
    B       Binary file
    G       Generated file
    N       Notice file
    S       Standard file
    U       Unknown file.

/ILoggerFactory.java
  S         text/x-java-source    ISO-8859-1
    MIT      MIT           The MIT License

/Image.png
  B         image/png

/LICENSE
  N         text/plain

/NOTICE
  N         text/plain

/Source.java
 !S         text/x-java-source    ISO-8859-1
    ?????    ?????         Unknown license (Unapproved)

```

So the '!' is on the line following the file name.  Note that file names can be very long, particularly when provided by the --input-source option.

I wanted to keep the filename as the first entry in a block, and I wanted to distinguish the mime type and encoding from the licenses so that it would be fairly easy to scan the output.

We could also expand the character prefix to the description or name.  For example `/Source.java` would look like

```
/Source.java
 !Standard file         text/x-java-source    ISO-8859-1
    ?????    ?????         Unknown license (Unapproved)
```
or
```
/Source.java
 !Standard         text/x-java-source    ISO-8859-1
    ?????    ?????         Unknown license (Unapproved)
```

but that looks a bit busy.

I could put the "! " at the beginning of the files that have problems

```
/ILoggerFactory.java
  S         text/x-java-source    ISO-8859-1
    MIT      MIT           The MIT License

! /Source.java
  S         text/x-java-source    ISO-8859-1
    ?????    ?????         Unknown license (Unapproved)
```

As a final suggestion in this response we could intend the file names by 2 spaces and leave the "!" hanging out there to be quickly found

```
  /ILoggerFactory.java
  S         text/x-java-source    ISO-8859-1
    MIT      MIT           The MIT License

  /Image.png
  B         image/png

  /LICENSE
  N         text/plain

  /NOTICE
  N         text/plain

! /Source.java
  S         text/x-java-source    ISO-8859-1
    ?????    ?????         Unknown license (Unapproved)
```

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
Claudenw: I will change it to prefix.

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
ottlinger: There seems to be a misunderstaning - I only had the feeling that the sentence in the report itself can be simplified to be more concise/clearer. Nothing more - no functional change or change in the report functionality.

Claudenw: The sentence was changed because the format changed and the "!" was no longer on the line with the file name.
I am not a report design person, not my wheel house.  If you have any suggestions that you think will make the report easier to read, scan and understand I would appreciate it.

I will be taking what we decide here to build the HTML version.

PR #406, opened 2024-11-28
RAT-405: Remove Sample from XML output

Claudenw:
Fix for RAT-405

removed SAMPLE collection and reporting


ottlinger:
When I first thought about this PR I wanted it to be configurable, but in bigger projects we should just let the user use their tooling to edit/check/verify the reported files.

ottlinger (approved):
Let us see how much this sppeds up the RAT runs ... thanks.

PR #409, opened 2024-12-08
RAT-98 report skipped files

Claudenw:
Fix for RAT-98

Type GENERATED has  been removed
Type IGNORED has been added

Formerly GENERATED files are now listed as IGNORED

The the optional boolean attribute isDirectory was added to the report XSD for resource nodes.

### XML report

Ignored files are added to the XML report as follows:

All ignored files have the isDirectory attribute set.

Ignored directories have the type set to IGNORED they are reported with a mediaType of application/octet-stream.  No file within or below the ignored directory is reported.

Files ignored by rule have the type set to IGNORED with the mediaType set but no encoding is specified

Files ignored because they are generated have the type set to IGNORED with the mediaType and encoding set as expected for the file.

**file ignored by rule**
```
<resource isDirectory="false" mediaType="text/plain" name="/ignored.file" type="IGNORED"/> 
```

**Ignored directory**
```
<resource isDirectory="true" mediaType="application/octet-stream" name="/ignored" type="IGNORED"/> 
```

** generated file **
```
<resource encoding="ISO-8859-1"  isDirectory="false" mediaType="text/plain" name="/generated.file" type="IGNORED"/> 
```

### Text report

**file ignored by rule**
```
  /ignored.file
  I          text/plain
```

**Ignored directory**
```
  /ignored
  I          application/octet-stream     (directory)
```

**generated file**
```
  /generated.file
  I          plain/text    ISO-8859-1
```

### Additional Changes 

- Moved /elements test resources into a new module `apache-rat-testdata` and used it to generate the example output for module `apache-rat`
- Removed junit and other testing libraries from the `apache-rat` combined jar.
-


Claudenw:
This got much larger than I wanted but a lot of it seems to be brittle tests dependant upon the XSLT output.  Part of the change here is to make the discovered brittle tests depend on the XML and not the output of the XSLT.

apache-rat-core/src/it/resources/ReportTest/RAT_335/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        <groupId>org.apache.rat</groupId>
+        <artifactId>apache-rat-plugin</artifactId>
+        <version>@pom.version@</version>
+        <configuration>
+          <outputStyle>xml</outputStyle>
+          <!-- Minimize the number of active rules to keep the test minimal -->
+          <excludes>
+            <exclude>**/.gitignore</exclude>
+          </excludes>
+          <useDefaultExcludes>false</useDefaultExcludes>
```
ottlinger, line 35: Should we use the new configuration elements here? (in order to provide examples of how to do it and dog-fooding ourselves?

apache-rat-core/src/it/resources/ReportTest/RAT_335/verify.groovy
```diff
@@ -0,0 +1,130 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+output = new File(args[0])
+content = output.text
+
+//Map<ClaimStatistic.Counter, String> data = new HashMap<>();
```
ottlinger, line 41: Should we remove the commented out stuff or do you need it for the SCM/GitIgnore things?

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
```diff
@@ -112,7 +112,7 @@
    char         type
     A       Archive file
     B       Binary file
-    G       Generated file
+    I       Ignored file
```
ottlinger, line 115: While having a look at this file I stumbled upon "missing-headers.xsl" in the same directory. The file does not seem to be used anymore - should we remove it as the functionality of integrating snippets of "marked" files was removed, @Claudenw ?

ottlinger: Having a second look it seems to referenced from multiple site-related files ..... so probably not safe to remove as it is an example transformation/report.

apache-rat-plugin/src/test/java/org/apache/rat/mp/RatCheckMojoTest.java
```diff
@@ -265,35 +330,103 @@ public void testRAT_343() throws Exception {
...
-                ReporterTestUtils.documentOut(false, Document.Type.STANDARD, "/dir3/file3.log") +
-                        ReporterTestUtils.UNKNOWN_LICENSE,
-        };
+//        final String[] expected = {
+//                ReporterTestUtils.counterText(ClaimStatistic.Counter.NOTICES, 1, false),
+//                ReporterTestUtils.counterText(ClaimStatistic.Counter.BINARIES, 0, false),
+//                ReporterTestUtils.counterText(ClaimStatistic.Counter.ARCHIVES, 0, false),
+//                ReporterTestUtils.counterText(ClaimStatistic.Counter.STANDARDS, 6, false),
+//                ReporterTestUtils.counterText(ClaimStatistic.Counter.IGNORED, 0, false),
+//                ReporterTestUtils.counterText(ClaimStatistic.Counter.UNKNOWN, 4, false),
```
ottlinger, line 342: Is this all due to RAT-355? @Claudenw

ottlinger (commented):
LGTM - so far. Thanks for all your efforts.

Is the commented out code due to the bug you found during implementation?

apache-rat-core/src/it/resources/ReportTest/RAT_335/pom.xml
Claudenw: Indeed.  I will make those changes in the next few days.

apache-rat-core/src/main/resources/org/apache/rat/plain-rat.xsl
Claudenw: missing-headers is one of the standard stylesheets listed in the help.  see https://github.com/apache/creadur-rat/blob/master/apache-rat-core/src/main/java/org/apache/rat/commandline/StyleSheets.java

apache-rat-core/src/it/resources/ReportTest/RAT_335/verify.groovy
Claudenw: I will remove in the next few days.

apache-rat-core/src/it/resources/ReportTest/RAT_335/pom.xml
Claudenw: The tests for RAT_355 should remain in their current state.  They will be replace with correct tests when 355 is corrected.  Hopefully by end of next week.

Claudenw:
@ottlinger  If you are OK with this please merge it.

The weird commented out code is because the the issue with RAT-355.  I hope to have it fixed in another PR by the end of next week.

PR #413, opened 2024-12-10
RAT-469: Fixes to correctly detect additional valid licenses

Claudenw:
A regression was spotted when another valid license was configured ..... but not properly reported.

If you configure GPL to be a valid license the recognition failed to pass.

### DIFF
--- a/apache-rat-tools/src/main/java/org/apache/rat/tools/Naming.java
+++ b/apache-rat-tools/src/main/java/org/apache/rat/tools/Naming.java
@@ -24,7 +24,7 @@ import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.io.Writer;
-import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Deque;
@@ -151,7 +151,7 @@ public final class Naming {
             };
         }
 
-        try (Writer underWriter = cl.getArgs().length != 0 ? new FileWriter(cl.getArgs()[0]) : new OutputStreamWriter(System.out, Charset.forName("UTF-8"));) {
+        try (Writer underWriter = cl.getArgs().length != 0 ? new FileWriter(cl.getArgs()[0]) : new OutputStreamWriter(System.out, StandardCharsets.UTF_8)) {
             if (cl.hasOption(CSV)) {
                 printCSV(columns, filter, cl.hasOption(CLI), mavenFilter, antFilter, descriptionFunction, underWriter);
             }


ottlinger:
Thanks for raising this issue!

Should we replace all comments in the Mojo as 
"@deprecated Use &lt;config&gt;." seems to be wrong as the configuration options appear directly under configuration, which is the standard element in Maven plugin configuration.

WDYT?

ottlinger:
@Claudenw apart from the fact the IT needs to be adapted once the patch is applied I wonder why GPL is not yet recognized and logged explicitly.

ottlinger:
@Claudenw I added GPL as a valid license - I always thought this could be configured and I do not need to add a new defaults.xml file in my project.

Claudenw:
I think I mistyped.  I said "So you have to add to the defaults" when what I meant was you need to add GPL-1, GPL-2, or GPL-3 as an approved license ID or add GPL as an approved license family.  The licenses are defined but they are not listed approved.

Claudenw:
> Should we replace all comments in the AbstractRatMojo as
> "@deprecated Use <config>." seems to be wrong as the configuration options appear directly under configuration, which is the standard element in Maven plugin configuration.

I think we need to step through each of them and make a decision.  "<config>" is not always wrong but there are better solutions for most entries.

ottlinger:
If I have a look at the log of the integration test I see:

```
[WARNING] Duplicate LicenseFamily category: AL    (action: IGNORE)
[WARNING] Duplicate LicenseFamily category: BSD-3 (action: IGNORE)
[WARNING] Duplicate LicenseFamily category: CDDL1 (action: IGNORE)
[WARNING] Duplicate LicenseFamily category: GEN   (action: IGNORE)
[WARNING] Duplicate LicenseFamily category: GPL   (action: IGNORE)
[WARNING] Duplicate LicenseFamily category: MIT   (action: IGNORE)
[WARNING] Duplicate LicenseFamily category: OASIS (action: IGNORE)
[WARNING] Duplicate LicenseFamily category: W3C   (action: IGNORE)
[WARNING] Duplicate LicenseFamily category: W3CD  (action: IGNORE)
```

could this be the cause for not recognizing GPL but labeling it UNKNOWN?
(apache-rat-plugin/target/invoker/RAT-469/build.log)

Claudenw:
First GPL should be recognized as a license FAMILY  it is in the default.xml
GPL-1, GPL-2, and GPL-3 should be recognised as license IDs also in the default.xml

```
[WARNING] Duplicate LicenseFamily category: W3C   (action: IGNORE)
[WARNING] Duplicate LicenseFamily category: W3CD  (action: IGNORE)
```
This indicates that at least 2 configurations were read and that the LicenseFamilies were defined in both.  I don't see how this happens but the warning is there so that if you accidentally overwrite an existing license you get warned.

ottlinger:
> First GPL should be recognized as a license FAMILY it is in the default.xml GPL1, GPL2, and GPL3 should be recognised as license IDs also in the default.xml
> 
> ```
> [WARNING] Duplicate LicenseFamily category: W3C   (action: IGNORE)
> [WARNING] Duplicate LicenseFamily category: W3CD  (action: IGNORE)
> ```
> 
> This indicates that at least 2 configurations were read and that the LicenseFamilies were defined in both. I don't see how this happens but the warning is there so that if you accidentally overwrite an existing license you get warned.

Just wanted to make sure that the IGNORE does not mean that any such license will be ignored.

ottlinger:
I merged in master and HeaderCheckWorkerTest is broken. 

@Claudenw Is it intended that the license family "unknown" is approved by default?

```
+++ b/apache-rat-core/src/test/java/org/apache/rat/analysis/HeaderCheckWorkerTest.java
@@ -43,8 +43,8 @@ public class HeaderCheckWorkerTest {
         ILicense matcher = new TestingLicense("test", "test");
         HeaderCheckWorker worker = new HeaderCheckWorker(new TestingMatcher(), new StringReader(""), Lists.list(matcher), subject);
         worker.read();
-        assertThat(subject.getMetaData().unapprovedLicenses().count()).isEqualTo(1);
-        assertThat(subject.getMetaData().unapprovedLicenses().collect(Collectors.toList()).get(0).getLicenseFamily()).isEqualTo(ILicenseFamily.UNKNOWN);
+        assertThat(subject.getMetaData().approvedLicenses().count()).isEqualTo(1);
+        assertThat(subject.getMetaData().approvedLicenses().collect(Collectors.toList()).get(0).getLicenseFamily()).isEqualTo(ILicenseFamily.UNKNOWN);
     }
```

Claudenw:
In this case "IGNORE" is the action taken.  It is configurable in software
but not config.

Claudenw:
Nope.  Unknown lice se can not be approved unless specifically added by the
user.  I will check on this this weekend

ottlinger:
> > Should we replace all comments in the AbstractRatMojo as
> > "@deprecated Use ." seems to be wrong as the configuration options appear directly under configuration, which is the standard element in Maven plugin configuration.
> 
> I think we need to step through each of them and make a decision. "" is not always wrong but there are better solutions for most entries.
> 
> I think this falls under the documentation verification step that we should get to in mid Jan.

I changed the default text in ef5c93bcc36d5f1e243179a283aad0fce13705e4

apache-rat-core/src/test/java/org/apache/rat/analysis/PolicyTest.java
```diff
@@ -0,0 +1,171 @@
+///*
```
ottlinger, line 1: Shall we remove this or enable it?

apache-rat-core/src/test/java/org/apache/rat/license/LicenseSetFactoryTest.java
```diff
@@ -0,0 +1,193 @@
...
+
+        licenseSetFactory.removeLicenseId(licenseId);
+        assertThat(licenseSetFactory.getApprovedLicensePredicate().test(license))
+                .as("Found unapproved id " + licenseId)
+                .isFalse();
+
+    }
+
+//        @Test
+//        public void testNonStandardDocumentsDoNotFailLicenseTests() {
```
ottlinger, line 173: Should these test cases be added somewhere else or just remove them?

apache-rat-core/src/test/java/org/apache/rat/analysis/PolicyTest.java
ottlinger: @Claudenw ping

Claudenw: I will have to make another pass to verify the tests are covered in more reasonable places.

Claudenw: I reworked the tests to ensure that they are still passing.

apache-rat-core/src/main/java/org/apache/rat/license/LicenseSetFactory.java
```diff
@@ -338,7 +344,7 @@ public SortedSet<String> getLicenseCategories(final LicenseFilter filter) {
                 return result;
             case APPROVED:
                 approvedLicenseCategories.stream().filter(s -> !removedLicenseCategories.contains(s)).forEach(result::add);
-                licenses.stream().filter(approved).forEach(l -> result.add(l.getLicenseFamily().getFamilyCategory()));
+                //licenses.stream().filter(approved).forEach(l -> result.add(l.getLicenseFamily().getFamilyCategory()));
```
ottlinger, line 347: @Claudenw is this a leftover or a TODO ?

ottlinger (approved):
Thanks for having another deeper look into the matter and for fixing the tests.

apache-rat-core/src/main/java/org/apache/rat/license/LicenseSetFactory.java
ottlinger: I removed the piece of code via [16a2319](https://github.com/apache/creadur-rat/pull/413/commits/16a2319181f4a60f414f04e42ab1932ec94207f6)

ottlinger:
GPL content is still recognised as UNKNOWN even if configured as an approvedLicense.

PR #417, opened 2024-12-26
RAT-98: Updated tests for gitignore changes & deprecated options

Claudenw:
Fixes issues raised by report skipped processing, specifically issues with Maven not correctly skipping files that should be skipped.

In the process of working on the bugs additional testing code was implemented to allows testing on file systems that are not native.  Most name based code now checks against UNIX, WINDOWS, and OSX.

Significant additional work was undertaken to ensure that the document names work as expected.

This is the last PR in the set and enables the tests for the gitignore changes as well as replaces deprecated options in some tests.


ottlinger:
@Claudenw did you accidentally push an old branch as the build fails with spotbugs errors?

Claudenw:
@ottlinger 
Actually, it is a work in progress.  I had to rework the descending exclusion processing because it was returning incorrect results.  That triggered a number of changes in integration tests and the like because they were not set correctly to begin with.  Net result is a large change but accurate results and more use of the assertJ package.

My next step is to see how I can split this into smaller pieces.

ottlinger (approved):
Due to the repo&branch mixup - LGTM.

Proposed changelog entry is:

```
      <action issue="RAT-98" type="add" dev="claudenw">
        Report skipped/excluded files and integrate testing of inclusion-exclusion and DocumentNames to work under Unix, Windows and Mac.
      </action>
```

Pls go ahead @Claudenw & thanks for your work to keep track of these changes.

PR #421, opened 2025-01-12
RAT-98: Added FSInfo to handle file system differences

Claudenw:
This a a part of #417 and is here to simplify the review.

The FSInfo allows us to easily test multiple file systems without having to create actual file systems.  It extracts away the file system info that we need for document name checks and add Google JimFS for testing.

Other changes include change DocumentName.baseName to a DocumentName rather than a string.  This adds some complexity to the builder but makes other checks much easier.

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/api/Document.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/api/Document.java
@@ -52,7 +52,7 @@ public abstract class Document implements Comparable<Document> {
     }
 
     /** The path matcher used by this document */
-    protected final DocumentNameMatcher nameExcluder;
+    protected final DocumentNameMatcher nameMatcher;
     /** The metadata for this document */
     private final MetaData metaData;
     /** The fully qualified name of this document */
@@ -61,11 +61,11 @@ public abstract class Document implements Comparable<Document> {
     /**
      * Creates an instance.
      * @param name the native NameSet of the resource.
-     * @param nameExcluder the document name matcher to filter directories/files.
+     * @param nameMatcher the document name matcher to filter directories/files.
      */
-    protected Document(final DocumentName name, final DocumentNameMatcher nameExcluder) {
+    protected Document(final DocumentName name, final DocumentNameMatcher nameMatcher) {
         this.name = name;
-        this.nameExcluder = nameExcluder;
+        this.nameMatcher = nameMatcher;
         this.metaData = new MetaData();
     }
 
@@ -81,8 +81,8 @@ public abstract class Document implements Comparable<Document> {
      * Gets the file filter this document was created with.
      * @return the file filter this document was created with.
      */
-    public final DocumentNameMatcher getNameExcluder() {
-        return nameExcluder;
+    public final DocumentNameMatcher getNameMatcher() {
+        return nameMatcher;
     }
 
     @Override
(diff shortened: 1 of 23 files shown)


Claudenw:
@ottlinger this change helps with Windows testing on Linux and visa-versa.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/plexus/MatchPatterns.java
```diff
@@ -41,15 +42,15 @@ private MatchPatterns(final MatchPattern[] patterns) {
 
     @Override
     public String toString() {
-        return source();
+        return Arrays.stream(patterns).map(MatchPattern::toString).collect(Collectors.toList()).toString();
```
ottlinger, line 45: Should we add a separator between entries to improve readability?

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/plexus/MatchPattern.java
```diff
@@ -116,7 +117,7 @@ public boolean startsWith(final String string) {
 
     @Override
     public String toString() {
-        return source;
+        return Arrays.asList(tokenized).toString();
```
ottlinger, line 120: Should we add a separator between entries to improve readability?

apache-rat-core/src/main/java/org/apache/rat/document/DocumentNameMatcher.java
```diff
@@ -219,4 +314,225 @@ public static DocumentNameMatcher and(final Collection<DocumentNameMatcher> matc
...
+    static class MatcherPredicate extends CollectionPredicateImpl {
+        MatcherPredicate(final Iterable<DocumentNameMatcher> matchers) {
+            super(matchers);
+        }
+
+        @Override
+        public boolean test(final DocumentName documentName) {
+            Iterator<DocumentNameMatcher> iter = getMatchers().iterator();
+            // included
+            if (iter.next().matches(documentName)) {
```
ottlinger, line 482: Is it correct to call iter.next() twice here or does that unintentionally skip the underlying iterator twice instead of a stepwise iteration OR is this a performance trick I just did not know about to iterate in a faster fashion?

ottlinger (changes requested):
Apart from my questions - LGTM

apache-rat-core/src/main/java/org/apache/rat/api/Document.java
```diff
@@ -61,11 +61,11 @@ public enum Type {
     /**
      * Creates an instance.
      * @param name the native NameSet of the resource.
-     * @param nameExcluder the document name matcher to filter directories/files.
+     * @param nameMatcher the document name matcher to filter directories/files.
      */
-    protected Document(final DocumentName name, final DocumentNameMatcher nameExcluder) {
+    protected Document(final DocumentName name, final DocumentNameMatcher nameMatcher) {
```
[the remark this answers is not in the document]
Claudenw: I was overzealous in the rename.  This actually is a matcher usage not excluder.

apache-rat-core/src/main/java/org/apache/rat/document/DocumentNameMatcher.java
Claudenw: It is correct.  I have cleaned it up to make more sense in PR #424

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/plexus/MatchPatterns.java
ottlinger: Sorry misunderstanding: I did not mean reformatting of the code, but a possible separator between distinct match patterns .....

ottlinger: Array#toString delivers this separator! resolved.

PR #422, opened 2025-01-12
RAT-98: changed 'name matcher' to 'document excluder'

Claudenw:
Changed name of variable(s) and method to be lexically correct.  It is not a document name matcher but rather an excluder.  

Part of #417 simplification.

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/OptionCollection.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/OptionCollection.java
@@ -249,10 +249,10 @@ public final class OptionCollection {
             DefaultLog.getInstance().error("Directory '" + documentName + "' does not exist.");
             return null;
         }
-        DocumentNameMatcher documentNameMatcher = config.getNameMatcher(documentName);
+        DocumentNameMatcher documentExcluder = config.getDocumentExcluder(documentName);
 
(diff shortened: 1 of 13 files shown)


ottlinger:
@Claudenw when I read the description I thought of "FilterEntrySkipper" ;) Not sure if DocumentExcluder is an easier2understand name. Should the class be renamed as well since DocumentNameMatcher seems rather "abstract" .... if it skips entries in the filtering/scanning process.

Claudenw:
> @Claudenw when I read the description I thought of "FilterEntrySkipper" or "DocumentSkipper/FilterSkipper" ;) Not sure if DocumentExcluder is an easier2understand name. Should the class be renamed as well since DocumentNameMatcher seems rather "abstract" .... if it skips entries in the filtering/scanning process.

@ottlinger 

The DocumentNameMatcher is untouched in this PR.  It does what it says.  It matches document names.

The excluder is a use of the document matcher.  That is documents that match it are excluded.  When reading the code it begins to make more sense and when debugging it is much easier to remember that if the name matches the excluder it is excluded.

I am open to other names but would prefer not to rename the DocumentNameMatcher in this PR.

ottlinger (approved):
LGTM - go ahead and merge @Claudenw

PR #425, opened 2025-01-14
RAT-98: Add file arg converter

Claudenw:
Added file converter to argument processing


ottlinger (approved):
LGTM - minor changes applied.

PR #426, opened 2025-01-14
RAT-471 - Publish build scans to develocity.apache.org

clayburn:
This PR migrates the RAT project to publish Build Scans to the the new Develocity instance at develocity.apache.org.

Additionally, this PR migrates sets a projectId for use by Develocity.

https://issues.apache.org/jira/browse/RAT-471

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -3,7 +3,7 @@
   <extension>
     <groupId>com.gradle</groupId>
     <artifactId>develocity-maven-extension</artifactId>
-    <version>1.23</version>
+    <version>1.22.2</version>
   </extension>
   <extension>
     <groupId>com.gradle</groupId>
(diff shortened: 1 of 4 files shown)


.mvn/extensions.xml
```diff
@@ -3,7 +3,7 @@
   <extension>
     <groupId>com.gradle</groupId>
     <artifactId>develocity-maven-extension</artifactId>
-    <version>1.23</version>
+    <version>1.22.2</version>
```
ottlinger, line 6: @clayburn why do we need a downgrade here?

clayburn: The current version of the server `2024.2.7` does not support version `1.23` of the extension. I think it happened to work, but strictly it is not supported until the Develocity server is on 2024.3.x.

Once this migration is complete to develocity.apache.org, upgrades should happen more quickly such that updating the extension before the server is not a concern.

PR #433, opened 2025-02-02
RAT-473: take global gitignore into account 

raboof:
Take into account exclusions from the global gitignore file if it is present in the default location. It's kind of icky to read environment variables in code called from an enum constructor, but I didn't see a more elegant solution yet either. Relatedly, I have not yet created a unit test (or made sure that a global gitignore will not interfere with a global gitignore being present), hence the WIP status.

https://git-scm.com/docs/gitignore

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
@@ -89,9 +89,11 @@ public enum StandardCollection {
                     "**/.project", "**/.settings/**", "**/.externalToolBuilders"),
             null, null),
     /**
-     * The files and directories created by GIT source code control to support GIT, also processes files listed in '.gitignore'.
+     * The files and directories created by GIT source code control to support GIT, also processes files listed in '.gitignore'
+     * and (unless RAT_NO_GIT_GLOBAL_IGNORE is specified) the global gitignore.
      */
-    GIT("The files and directories created by GIT source code control to support GIT, also processes files listed in '.gitignore'.",
+    GIT("The files and directories created by GIT source code control to support GIT, also processes files listed in '.gitignore' " +
+        "and (unless RAT_NO_GIT_GLOBAL_IGNORE is specified) the global gitignore.",
             Arrays.asList("**/.git/**", "**/.gitignore"),
             null,
             new GitIgnoreBuilder()
--- a/apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/AbstractFileProcessorBuilder.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/AbstractFileProcessorBuilder.java
@@ -203,10 +203,14 @@ public abstract class AbstractFileProcessorBuilder {
         return result == null ? new File[0] : result;
     }
 
+    protected LevelBuilder getLevelBuilder(final int level) {
+        return levelBuilders.computeIfAbsent(level, k -> new LevelBuilder());
+    }
+
     /**
      * Manages the merging of {@link MatcherSet}s for the specified level.
      */
-    private static final class LevelBuilder {
+    protected static final class LevelBuilder {
         /**
          * The list of MatcherSets that this builder produced.
          */
--- a/apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilder.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilder.java
@@ -19,7 +19,12 @@
 package org.apache.rat.config.exclusion.fileProcessors;
 
 import java.io.File;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Consumer;
 
 import org.apache.rat.config.exclusion.ExclusionUtils;
@@ -51,6 +56,41 @@ public class GitIgnoreBuilder extends AbstractFileProcessorBuilder {
      */
     public GitIgnoreBuilder() {
         super(IGNORE_FILE, COMMENT_PREFIX, true);
+    }
+
+    private MatcherSet processGlobalIgnore(final Consumer<MatcherSet> matcherSetConsumer, final DocumentName root, final DocumentName globalGitIgnore) {
+        final MatcherSet.Builder matcherSetBuilder = new MatcherSet.Builder();
+        final List<String> iterable = new ArrayList<>();
+        ExclusionUtils.asIterator(globalGitIgnore.asFile(), commentFilter)
+                .map(entry -> modifyEntry(matcherSetConsumer, globalGitIgnore, entry).orElse(null))
+                .filter(Objects::nonNull)
+                .map(entry -> ExclusionUtils.qualifyPattern(root, entry))
+                .forEachRemaining(iterable::add);
+
+        Set<String> included = new HashSet<>();
+        Set<String> excluded = new HashSet<>();
+        MatcherSet.Builder.segregateList(excluded, included, iterable);
+        DocumentName displayName = DocumentName.builder(root).setName("global gitignore").build();
+        matcherSetBuilder.addExcluded(displayName, excluded);
+        matcherSetBuilder.addIncluded(displayName, included);
+        return matcherSetBuilder.build();
+    }
+
+    @Override
+    protected MatcherSet process(final Consumer<MatcherSet> matcherSetConsumer, final DocumentName root, final DocumentName documentName) {
+      if (root.equals(documentName.getBaseDocumentName())) {
+          Optional<File> globalGitIgnore = globalGitIgnore();
+          List<MatcherSet> matcherSets = new ArrayList<MatcherSet>();
+          matcherSets.add(super.process(matcherSetConsumer, root, documentName));
+          if (globalGitIgnore.isPresent()) {
+              LevelBuilder levelBuilder = getLevelBuilder(Integer.MAX_VALUE);
+              DocumentName ignore = DocumentName.builder(globalGitIgnore.get()).build();
+              matcherSets.add(processGlobalIgnore(levelBuilder::add, root, ignore));
+          }
+          return MatcherSet.merge(matcherSets);
+      } else {
+          return super.process(matcherSetConsumer, root, documentName);
+      }
     }
 
     /**
@@ -102,4 +142,34 @@ public class GitIgnoreBuilder extends AbstractFileProcessorBuilder {
         }
         return Optional.of(prefix ? NEGATION_PREFIX + pattern : pattern);
     }
+
+    /**
+     * The global gitignore file to process, based on the
+     * RAT_NO_GIT_GLOBAL_IGNORE, XDG_CONFIG_HOME, and HOME environment
+     * variables.
+     */
+    protected Optional<File> globalGitIgnore() {
+        if (System.getenv("RAT_NO_GIT_GLOBAL_IGNORE") != null) {
+            return Optional.empty();
+        }
+
+        String xdgConfigHome = System.getenv("XDG_CONFIG_HOME");
+        String filename;
+        if (xdgConfigHome != null && !xdgConfigHome.isEmpty()) {
+            filename = xdgConfigHome + File.separator + "git" + File.separator + "ignore";
+        } else {
+            String home = System.getenv("HOME");
+            if (home == null) {
+                home = "";
+            }
+            filename = home + File.separator + ".config" + File.separator + "git" + File.separator + "ignore";
+        }
+        File file = new File(filename);
+        if (file.exists()) {
+            return Optional.of(file);
+        } else {
+            return Optional.empty();
+        }
+    }
+
 }
--- a/apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilderTest.java
+++ b/apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilderTest.java
@@ -56,7 +56,12 @@ public class GitIgnoreBuilderTest extends AbstractIgnoreBuilderTest {
 
             writeFile(".gitignore", Arrays.asList(lines));
 
-            assertCorrect(new GitIgnoreBuilder(), matches, notMatches);
+            assertCorrect(new GitIgnoreBuilder() {
+                @Override
+                protected Optional<File> globalGitIgnore() {
+                    return Optional.empty();
+                }
+            }, matches, notMatches);
         } finally {
             System.getProperties().remove("FSInfo");
         }
@@ -106,7 +111,7 @@ public class GitIgnoreBuilderTest extends AbstractIgnoreBuilderTest {
             DocumentName name = documentName.resolve(test);
             assertThat(matcher.matches(name)).as(test).isTrue();
         }
-        for (String test: notMatching) {
+        for (String test : notMatching) {
             DocumentName name = documentName.resolve(test);
             assertThat(matcher.matches(name)).as(test).isFalse();
         }
@@ -114,7 +119,12 @@ public class GitIgnoreBuilderTest extends AbstractIgnoreBuilderTest {
 
     @Test
     public void test_RAT_335() {
-        GitIgnoreBuilder underTest = new GitIgnoreBuilder();
+        GitIgnoreBuilder underTest = new GitIgnoreBuilder() {
+            @Override
+            protected Optional<File> globalGitIgnore() {
+                return Optional.empty();
+            }
+        };
         URL url = GitIgnoreBuilderTest.class.getClassLoader().getResource("GitIgnoreBuilderTest/src/");
         File file = new File(url.getFile());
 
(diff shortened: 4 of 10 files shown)


apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
```diff
@@ -310,4 +310,18 @@ public boolean hasStaticDocumentNameMatcher() {
...
+        if (xdgConfigHome != null && !xdgConfigHome.isEmpty()) {
+            return xdgConfigHome + File.pathSeparator + "git" + File.pathSeparator + "ignore";
+        } else {
+            String home = System.getenv("HOME");
+            if (home == null) {
+                home = "";
+            }
+            return home + File.pathSeparator + ".config" + File.pathSeparator + "git" + File.pathSeparator + "ignore";
+        }
+    }
```
Claudenw, line 326: Have you tested this on a windows system?  I just got back from FOSDEM and am trying to clean up a bad merge so I can't check this right now but, I think that the slashes are required to be "/" in this pattern list.    There are methods that will convert them for you -- but again that is in the code I am trying to fix.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
```diff
@@ -310,4 +310,18 @@ public boolean hasStaticDocumentNameMatcher() {
         }
         return false;
     }
+
+    /** The location of the global gitignore file to process */
+    private static String globalGitIgnore() {
+        String xdgConfigHome = System.getenv("XDG_CONFIG_HOME");
+        if (xdgConfigHome != null && !xdgConfigHome.isEmpty()) {
+            return xdgConfigHome + File.pathSeparator + "git" + File.pathSeparator + "ignore";
```
Claudenw, line 318: Are you certain that the default GIT process is to ignore `XDG_CONFIG_HOME` if it is empy?  An empty `XDG_CONFIG_HOME` could be a way to ignore the `HOME` based isgnore.  Please verify.

Claudenw (changes requested):
Just a couple of questions.

ottlinger:
@raboof would you mind creating a RAT-jira ticket to describe your intentions? Thanks

Do you have a use case where I have something in my global gitignore that is not part of my project? It seems a bit mysterious to me. Thx for a bit more context.

raboof:
> @raboof would you mind creating a [RAT-jira](https://issues.apache.org/jira/browse/RAT-jira) ticket to describe your intentions? Thanks
> 
> Do you have a use case where I have something in my global gitignore that is not part of my project? It seems a bit mysterious to me. Thx for a bit more context.

Done (RAT-473). Yes, I put some things in there for development tools that I use and that use files in the project directory. That way I don't need to have a perpetually-dirty .gitignore in the project or contribute those unusual special cases to the upstream .gitignore.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/StandardCollection.java
ottlinger: Would it make sense to provide this as a separate configuration option? - I wonder if it takes too much time to find the global ignore file on all platforms/operating systems upon each run. WDYT?

raboof: > Would it make sense to provide this as a separate configuration option?

I'll have a look if/how that fits in the new structure Claude is preparing. I think it'd make sense to follow what git does by default, but having a way to suppress it could be good.

> I wonder if it takes too much time to find the global ignore file on all platforms/operating systems upon each run. WDYT?

That sounds unlikely to me: finding it is only 1-2 environment variable lookups and opening 1 file. Given RAT then proceeds to open *every* unignored file in the repo, the extra time seems negligible?

Claudenw:
@raboof Are some of the things you put in your private ignore file found in other StandardCollection entries?  Would it make sense to create more StandardCollection entries for anything that is missing?

The changes that you want to work against are in Pull request #439 
Specifically:

-  in the `AbstractFileProcessorBuilder.build(final DocumentName root)`  you want to add a call to a new method `void initializeBuilder(final DocumentName root)` that you will create in `AbstractFileProcessorBuilder` as a method that does nothing.
- in `GitIgnoreBuilder` you want to implement `initializeBuilder()` to perform the file name lookup logic and if a file is found process it using by createing a `LevelBuilder` for level `-1` and calling `protected MatcherSet process(final Consumer<MatcherSet> matcherSetConsumer, final DocumentName root, final DocumentName documentName)` to process the file. 

See `AbstractFileProcessBuilder.checkDirectory()` for an example of creating and using the LevelBuilder.

Bascically the system works by creating `MatcherSet`s for each level (.gitignore) file and then applies them in reverse order so that the later ones override the earlier ones as is requried by the gitignore spec.  A `MatcherSet` has two matchers, one that specifies the files that are  to be _explicitly_ included (overrides earlier excludes), and one to list files to exclude.  The `process()` method will do that correctly if you have the proper `root` and `documentName` specified.

Claudenw:
@raboof all code changes are now in master.  Please feel free to build against that.  There are more changes coming but they are mostly testing and should not impact your changes.

raboof:
Great, thanks for the pointers! This will likely be on the backburner for a while but I'll return to it eventually :)

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/AbstractFileProcessorBuilder.java
```diff
@@ -151,7 +161,7 @@ protected MatcherSet process(final Consumer<MatcherSet> matcherSetConsumer, fina
         ExclusionUtils.asIterator(documentName.asFile(), commentFilter)
                 .map(entry -> modifyEntry(matcherSetConsumer, documentName, entry).orElse(null))
                 .filter(Objects::nonNull)
-                .map(entry -> ExclusionUtils.qualifyPattern(documentName, entry))
+                .map(entry -> ExclusionUtils.qualifyPattern(root, entry))
```
raboof, line 164: this used to assume the pattern is relative to the location of the `.gitignore` file, which used to be true, but is not the case for the global gitignore.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/AbstractFileProcessorBuilder.java
```diff
@@ -166,18 +176,17 @@ protected MatcherSet process(final Consumer<MatcherSet> matcherSetConsumer, fina
...
      * @param directory The name of the directory to process.
      * @param fileFilter the filter to detect processable files with.
      */
-    private void checkDirectory(final int level, final DocumentName root, final DocumentName directory, final FileFilter fileFilter) {
+    private void checkDirectory(final int level, final DocumentName directory, final FileFilter fileFilter) {
         File dirFile = directory.asFile();
         for (File file : listFiles(dirFile, fileFilter)) {
             LevelBuilder levelBuilder = levelBuilders.computeIfAbsent(level, k -> new LevelBuilder());
-            levelBuilder.add(process(levelBuilder::add, root, DocumentName.builder(file).build()));
+            levelBuilder.add(process(levelBuilder::add, DocumentName.builder(file).build(), DocumentName.builder(file).build()));
```
raboof, line 186: for the non-global gitignore, this assumption has now moved here

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilder.java
```diff
@@ -53,6 +53,16 @@ public GitIgnoreBuilder() {
         super(IGNORE_FILE, COMMENT_PREFIX, true);
     }
 
+    @Override
+    void initializeBuilder(final DocumentName root) {
+        File globalGitIgnore = new File(globalGitIgnore());
+        if (globalGitIgnore.exists()) {
+            LevelBuilder levelBuilder = levelBuilders.computeIfAbsent(-1, k -> new LevelBuilder());
+            DocumentName documentName = DocumentName.builder(globalGitIgnore).build();
+            levelBuilder.add(process(levelBuilder::add, root, documentName));
```
raboof, line 62: for the global gitignore, we can pass the root of the project instead

apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilderTest.java
```diff
@@ -125,13 +135,50 @@ public void test_RAT_335() {
...
+        DocumentName documentName = DocumentName.builder(file).build();
+        List<MatcherSet> matcherSets = underTest.build(documentName);
+        DocumentNameMatcher matcher = MatcherSet.merge(matcherSets).createMatcher();
+
+        DocumentName candidate = DocumentName.builder()
+                .setName("/home/claude/apache/creadur-rat/apache-rat-core/target/test-classes/GitIgnoreBuilderTest/src/dir1/file1.log")
+                .setBaseName("home/claude/apache/creadur-rat/apache-rat-core/target/test-classes/GitIgnoreBuilderTest/src/").build();
+        System.out.println("Decomposition for " + candidate);
+
+        assertThat(matcher.toString()).isEqualTo("matcherSet(or('included .gitignore', 'included .gitignore'), or('excluded .gitignore', **/.gitignore, 'excluded .gitignore', 'excluded " + globalGitIgnore.substring(1) + "'))");
```
raboof, line 177: `globalGitIgnore.substring(1)` does not give the right presentation on Windows. I don't have easy access to a Windows machine for experimenting - could someone help figure out the right incantation here?

raboof (commented):
Updated on top of @Claudenw's refactor now, much neater!

One challenge remaining in the test, and we could still make this feature optional (but I'd really like to enable it by default).

apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilderTest.java
ottlinger: @raboof not sure if you see the test failure results, but it is:

```
Error:  Failures: 
Error:    GitIgnoreBuilderTest.test_global_gitignore:177 
expected: "matcherSet(or('included .gitignore', 'included .gitignore'), or('excluded .gitignore', **/.gitignore, 'excluded .gitignore', 'excluded D:/a/creadur-rat/creadur-rat/apache-rat-core/target/test-classes/GitIgnoreBuilderTest/global-gitignore'))"
 but was: "matcherSet(or('included .gitignore', 'included .gitignore'), or('excluded .gitignore', **/.gitignore, 'excluded .gitignore', 'excluded a/creadur-rat/creadur-rat/apache-rat-core/target/test-classes/GitIgnoreBuilderTest/global-gitignore'))"
```

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilder.java
```diff
@@ -102,4 +112,19 @@ protected Optional<String> modifyEntry(final Consumer<MatcherSet> matcherSetCons
         }
         return Optional.of(prefix ? NEGATION_PREFIX + pattern : pattern);
     }
+
+    /** The location of the global gitignore file to process */
+    protected String globalGitIgnore() {
```
Claudenw, line 117: This method should return an Optional<File>.  Check if any of the files exist and return the file otherwise return `Optional.empty()`  Your `initializationBuilder` implementation can then handle the case where the file exists more easily,.

Also, this will give us a single location to enable/disable the use of the global ignore.

Add a check for "RAT_NO_GIT_GLOBAL_IGNORE" environment var.  If it is set return `Optional.empty()`.

This will allow users to disable this feature if desired.  

You will also need to update the documentation for GIT in the StandardCollection enum.  It should mention that the global ignore is active and can be overridden by setting the environment variable "RAT_NO_GIT_GLOBAL_IGNORE"

Finally, if you detect RAT_NO_GIT_GLOBAL_IGNORE log an info entry indicating that global ignore is skipped because of RAT_NO_GIT_GLOBAL_IGNORE

apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilderTest.java
Claudenw: Create a document name from the gitignore file name:
```
DocumentName globalIgnoreName = DocumentName.builder(new File(globalGitIgnoreUrl.getFile())).build();
```

Then use `globalIgnoreName.localized("/").substring(1)`

When dealing with Windows/Linux file name differences `DocumentName` is your friend.

Claudenw (changes requested):
Sorry it has taken me soooo long to get to this PR.  I'll try to do better going forward.

apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilderTest.java
raboof: Thanks! That seems to have worked (though I did seem to need a `substring`)

raboof:
> Sorry it has taken me soooo long to get to this PR. I'll try to do better going forward.

No worries!

Claudenw (changes requested):
I had to sit down and go through this code with a debugger since I have not been thinking about this bit.  There is a lot here to think about.  I hope my comments have made it clear where things need to change.

There are 2 things to think about:

1. Order of test execution.  Include before exclude.  Items farthest down the  tree first, and global rules before anything.
2. Display name as seen in `DocumentNameMatcher.toString()`, this provides information about order and location of files that contribute to the inclusion/exclusion.

I look forward to your update.

raboof:
(I understand the problem with the previous implementation. I have mostly applied the suggested changes, but I don't fully understand all parts yet, and would like to have a closer look at both the implementation and the test - hence 'draft' status)

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilder.java
```diff
@@ -53,6 +58,43 @@ public GitIgnoreBuilder() {
...
+    }
+
+    @Override
+    protected MatcherSet process(final Consumer<MatcherSet> matcherSetConsumer, final DocumentName root, final DocumentName documentName) {
+      if (root.equals(documentName.getBaseDocumentName())) {
+          Optional<File> globalGitIgnore = globalGitIgnore();
+          List<MatcherSet> matcherSets = new ArrayList<MatcherSet>();
+          matcherSets.add(super.process(matcherSetConsumer, root, documentName));
+          if (globalGitIgnore.isPresent()) {
+              LevelBuilder levelBuilder = getLevelBuilder(-1);
```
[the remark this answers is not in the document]
raboof: Ah yes, sorry for not calling that out - I wanted to make sure this is covered in a testcase before making the code change :)

raboof:
> Include before exclude.

I see in the current implementation indeed includes take precedence over excludes, regardless of where they're specified. Is that really correct for Git? AFAICT a rule in the repo directory should take precedence over a rule in the global gitignore, regardless of whether it's an include or an exclude.

https://git-scm.com/docs/gitignore

raboof:
I see in the current implementation indeed includes (i.e. explicitly "not-ignored" patterns) take precedence over excludes (ignores), regardless of where they're specified. Is that really correct for Git? AFAICT a rule in the repo directory should take precedence over a rule in the global gitignore, regardless of whether it's an include or an exclude.

https://git-scm.com/docs/gitignore

Claudenw:
The RAT process assumes all files are included by default.  So the exclude files do just that list exceptions.  Then there is the exception to the exception, where a file is explicitly added back.  Once added back we do not support removing them again.  This is the basic process for RAT inclusion/exclusion.

In order to provide scoped include/exclude and precedence of more specificity over less specificity the process splits the include/exclude into 2 sets internally.  So if we have 3 layers of specificity we have 3 sets of include rules and 3 sets of exclude rules.  When we execute the rules we execute the include rules first in order of most specific to least specific set we then execute the exclude rules in the same order.  The first rule that fires determines how the files is treated.  If no rule fires than we include it (default).

As I write this, I do not have the code in front of me.  So the above may be slightly incorrect.

If label the includes as I1- I3 with I3 having the highest precedence, and similarly excludes as E1-E3.
I think we execute  I3, I2, I1, E3, E2, E1   But it may actually be coded as I3, E3, I2, E2, I1, E1.

Using the decomposition example I provided earlier we should be able to determine how it actually executes.   It also should not be too much work to put them in a different order.  If we need to change the ordering let's do that with a different issue and PR please

ottlinger:
@raboof would you mind adding a changelog entry to describe the newly added feature? thx
src/changes/changes.xml

apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilderTest.java
```diff
@@ -125,13 +135,56 @@ public void test_RAT_335() {
         DocumentName candidate = DocumentName.builder()
                 .setName("/home/claude/apache/creadur-rat/apache-rat-core/target/test-classes/GitIgnoreBuilderTest/src/dir1/file1.log")
                 .setBaseName("home/claude/apache/creadur-rat/apache-rat-core/target/test-classes/GitIgnoreBuilderTest/src/").build();
-        System.out.println("Decomposition for "+candidate);
+        System.out.println("Decomposition for " + candidate);
```
Claudenw, line 138: Remove these lines.  They are junk I missed before.

apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilderTest.java
```diff
@@ -125,13 +135,56 @@ public void test_RAT_335() {
...
+        File file = new File(url.getFile());
+
+        DocumentName documentName = DocumentName.builder(file).build();
+        List<MatcherSet> matcherSets = underTest.build(documentName);
+        DocumentNameMatcher matcher = MatcherSet.merge(matcherSets).createMatcher();
+
+        DocumentName candidate = DocumentName.builder()
+                .setName("/home/claude/apache/creadur-rat/apache-rat-core/target/test-classes/GitIgnoreBuilderTest/src/dir1/file1.log")
+                .setBaseName("home/claude/apache/creadur-rat/apache-rat-core/target/test-classes/GitIgnoreBuilderTest/src/").build();
+        System.out.println("Decomposition for " + candidate);
```
Claudenw, line 177: Remove these lines.  They are junk I missed deleting before.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/fileProcessors/GitIgnoreBuilder.java
```diff
@@ -102,4 +142,30 @@ protected Optional<String> modifyEntry(final Consumer<MatcherSet> matcherSetCons
         }
         return Optional.of(prefix ? NEGATION_PREFIX + pattern : pattern);
     }
+
+    /** The location of the global gitignore file to process */
```
Claudenw, line 146: Please expand this comment to indicate that the processing of RAT_NO_GIT_GLOBAL_IGNORE, XDG_CONFIG_HOME, and HOME environment variables.

Claudenw (changes requested):
Looks good to me.  Thank you for your patience and hard work.  I do have one more change to note.

Please update `StandardCollection`, the `GIT` entry to indicate that the global exclude file will be processed unless "RAT_NO_GIT_GLOBAL_IGNORE" environment variable is specified.  This will surface the information in the help system.

raboof:
> Thank you for your patience and hard work

Thanks for your in-depth review and extensive advice!

Claudenw (approved):
Thanks again.

Claudenw:
What is the difference?

LinkedIn: http://www.linkedin.com/in/claudewarren

PR #436, opened 2025-02-06
RAT-98: Improve logging messages

Claudenw:
Added/updated logging messages.

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/configuration/XMLConfigurationReader.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/configuration/XMLConfigurationReader.java
@@ -21,7 +21,6 @@ package org.apache.rat.configuration;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
-import java.io.StringWriter;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


apache-rat-core/src/main/java/org/apache/rat/configuration/XMLConfigurationReader.java
```diff
@@ -126,6 +127,21 @@ public SortedSet<ILicenseFamily> licenseFamilies() {
         };
     }
 
+    /**
+     * Creates textual representation of a node for display.
+     * @param node the node to create the textual representation of.
+     * @return The textual representation of the node for display.
+     */
+    private String nodeText(final Node node) {
+        StringWriter stringWriter = new StringWriter().append("<").append(node.getNodeName());
```
ottlinger, line 136: Should we use StringBuilder here instead?

ottlinger (changes requested):
Could we use StringBuilder instead of StringWriter as the methods are not part of streaming methods, but just log stuff.

apache-rat-core/src/main/java/org/apache/rat/configuration/XMLConfigurationReader.java
Claudenw: Changed

Claudenw: @ottlinger any other changes you want or may I merge?

ottlinger (approved):
Thanks, LGTM

PR #438, opened 2025-02-06
RAT-98: converted to assertj

Claudenw:
Convert to assertj

Normalize test directory names.

Ingore tests that will be refactored in later RAT-98 changes


ottlinger (approved):
LGTM, thanks.

PR #439, opened 2025-02-06
RAT-98:  Converted to FileProcessorBuilder design

Claudenw:
Changed from FileProcessor to FileProcessorBuilder to simplify processors that have multiple layers.


apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/AbstractIgnoreBuilderTest.java
```diff
@@ -0,0 +1,122 @@
...
+    @TempDir
+    protected Path tmpPath;
+    protected DocumentName baseName;
+
+    @BeforeEach
+    protected void setup() throws IOException {
+        baseName = DocumentName.builder(tmpPath.toFile()).build();
+    }
+
+    protected static Stream<Arguments> fsInfoArgs() {
```
ottlinger, line 59: Is this method needed /used in other PRs?

ottlinger (approved):
LGTM, except for the question about the unused method.

apache-rat-core/src/test/java/org/apache/rat/config/exclusion/fileProcessors/AbstractIgnoreBuilderTest.java
Claudenw: Turns out, no.  I removed it.

PR #442, opened 2025-02-08
RAT-98: updated DocumentNameMatcher

Claudenw:
Update DocumentNameMatcher

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/document/DocumentNameMatcher.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/document/DocumentNameMatcher.java
@@ -302,7 +302,7 @@ public final class DocumentNameMatcher {
         }
         List<DocumentNameMatcher> workingSet = Arrays.asList(includes, excludes);
         return new DocumentNameMatcher(format("matcherSet(%s)", join(workingSet)),
-                new CollectionPredicateImpl(workingSet) {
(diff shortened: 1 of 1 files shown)


apache-rat-core/src/main/java/org/apache/rat/document/DocumentNameMatcher.java
```diff
@@ -400,9 +399,8 @@ public String toString() {
     interface CollectionPredicate extends Predicate<DocumentName> {
         Iterable<DocumentNameMatcher> getMatchers();
     }
-
     /**
-     * A marker class to indicate this predicate contains a collection of matchers.
+     * CollectionPredicate implementation.
      */
     abstract static class CollectionPredicateImpl implements CollectionPredicate {
```
ottlinger, line 405: @Claudenw would DefaultCollectionPredicate be a better name for the implementation? or RatCollectionPredicate instead?

ottlinger (approved):
LGTM - just a minor naming question concerning the actual implemenation class.

apache-rat-core/src/main/java/org/apache/rat/document/DocumentNameMatcher.java
Claudenw: Switched to DefaultCollectionPredicate

PR #450, opened 2025-03-16
RAT-398: deprecate methods in Ant report

Claudenw:
Deprecated Ant methods and ensured that deprecation message appears in log.

Closes RAT-398


jbonofre (approved):

ottlinger:
We are still in the analysis why the tests fail and should not merge this PR until the problems are fixed.

apache-rat-plugin/pom.xml
```diff
@@ -160,7 +160,7 @@
             <configuration>
               <includeArtifactIds>apache-rat-core</includeArtifactIds>
               <excludeClassifiers>tests</excludeClassifiers>
-              <include>META-INF/RELEASE_NOTES.txt</include>
+              <includes>META-INF/RELEASE_NOTES.txt</includes>
```
ottlinger, line 163: @Claudenw should we add a configuration for maven-clean-plugin as well in order to get rid of the generated files?

apache-rat-tasks/pom.xml
```diff
@@ -80,6 +80,12 @@
       <artifactId>junit-jupiter-params</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
ottlinger, line 83: What is this dependency used for?

apache-rat-tasks/pom.xml
```diff
@@ -118,29 +124,12 @@
...
               <includeArtifactIds>apache-rat-core</includeArtifactIds>
               <excludeClassifiers>tests</excludeClassifiers>
-              <include>META-INF/RELEASE_NOTES.txt</include>
+              <includes>META-INF/RELEASE_NOTES.txt</includes>
               <outputDirectory>${project.build.outputDirectory}</outputDirectory>
             </configuration>
           </execution>
         </executions>
       </plugin>
-      <plugin>
```
ottlinger, line 127: We should check if the project still works in Eclipse. I remember that Eclipse sometimes does strange things with generated files that are copied into the source tree ....

apache-rat-tasks/src/main/java/org/apache/rat/anttasks/All.java
```diff
@@ -27,6 +28,7 @@
  * @deprecated use configuration file.
  */
 @Deprecated // since 0.17
+@DeprecationReporter.Info(since = "0.17", forRemoval = true, use = "Configuration file and <config> element")// since 0.17
```
ottlinger, line 31: Should this be <configuration> element here? Or is that only valid for the Maven plugin? Just to be correct in the help output ....

apache-rat-tasks/src/test/resources/helpTest/run.sh
```diff
@@ -0,0 +1 @@
+ant -lib /home/claude/apache/creadur-rat/apache-rat/target/apache-rat-0.17-SNAPSHOT.jar execHelp
```
ottlinger, line 1: Should this be checked in under .buildtools?

apache-rat-tasks/src/main/java/org/apache/rat/anttasks/All.java
ottlinger: This should be changed among many files in this PR once it's clear which element name is the correct one. I was unable to find any ANT examples that used a config, thus I'll keep my comment as a reminder here :)

apache-rat-tasks/pom.xml
ottlinger: I found the NotNull in the generated BaseAntTask - what is it good for in the generated file/what does it prevent me from doing?

ottlinger:
@jbonofre do you have an idea concerning the failing test? There were no changes in the test; it runs well on master, but fails in this branch. The junit-update did not bring in the problem as I checked locally.

@Claudenw over the weekend I thought maybe it is a timing issue and wanted to try to add a local call to GC as that fixed some of the other test failures before ..... really odd situation.

Claudenw:
@ottlinger @jbonofre I spent some time looking at this and discovered that there were a few junit4 annotations still in use.  The one I was most concerned about was the the `@Rule` annotation.  I removed all junit4 references and everything worked.

Claudenw:
@ottlinger @jbonofre does anyone see any reason not to merge this now?

ottlinger:
@Claudenw there are some other review comments ..... cool, that you seem to have fixed the jUnit problem!

apache-rat-tasks/src/test/resources/helpTest/run.sh
ottlinger: Should we remove this helper script? @Claudenw

Claudenw: It was just an example of how to execute the help from the command line.  I removed it.

apache-rat-tasks/src/main/java/org/apache/rat/anttasks/All.java
Claudenw: These files are in the current release.  We need to retain them for a release and mark them as deprecated so people have a heads up before we remove them in V1.

apache-rat-tasks/pom.xml
Claudenw: It helps the IDE but I have removed it.

ottlinger (approved):
Thanks for all your work, I filed RAT-474 concerning the Eclipse problems.

apache-rat-tasks/pom.xml
ottlinger: Filed RAT-474 to handle Eclipse problems separately (they happen on master and this feature branch)!

PR #478, opened 2025-06-14
RAT-379, RAT-475: Fix poms and module issues in order to start with RAT from a clean slate

Claudenw:
This change fixes some issues with building RAT starting from a clean slate.

Examples of commands that failed:

- `mvn test` failed on test code from clean build.
- `mvn compile` failed on core tests  if there were no RAT jars in the local `$HOME/.m2/repository` tree.

The issue was the system was looking for jars that were being built during the process so if the process had execute `mvn install` the build would run.  Also `mvn package` would work because it handles building the dependencies.

this change should mean that running `mvn compile` should work in any sub module directory after a clean. 

also removes dependency on `jwigrv` to move the RELEASE NOTES file.


Claudenw:
@ottlinger I think this change fixes most (all?) of the problems that were found in when starting from a clean build.  Please check the issues you encountered.

pom.xml
```diff
@@ -85,11 +86,11 @@ agnostic home for software distribution comprehension and audit tools.
         <artifactId>apache-rat-core</artifactId>
         <version>${project.version}</version>
       </dependency>
-      <dependency>
-        <groupId>org.apache.rat</groupId>
-        <artifactId>apache-rat-testdata</artifactId>
-        <version>${project.version}</version>
-      </dependency>
+<!--      <dependency>-->
```
ottlinger, line 89: Could we remove this?

pom.xml
```diff
@@ -856,7 +858,7 @@ agnostic home for software distribution comprehension and audit tools.
     <module>apache-rat-tasks</module>
     <module>apache-rat</module>
     <module>apache-rat-tools</module>
-    <module>apache-rat-testdata</module>
+<!--    <module>apache-rat-testdata</module>-->
```
ottlinger, line 861: This removes the whole submodule from the project? Is that intended?

Claudenw:
closing will not merge

PR #479, opened 2025-06-15
removed jar from core/src tree

Claudenw:


Claudenw:
Closed due to incorrect assumption

PR #480, opened 2025-06-15
fixed RELEASE_NOTES.txt distribution

Claudenw:


ottlinger (changes requested):
Seems fine: a local project and site build looked ok and the file is packaged within the resulting JAR - BUT
the problem reported in RAT-379 has returned while opening the project in Eclipse itself:
"Path must include project and resource name: /"

PR #481, opened 2025-06-15
removed testdata jar

Claudenw:
removed testdata jar, reworked poms to include data and moved data from main to test directory


ottlinger (changes requested):
I'm able to build the project, but an import into Eclipse causes -
MojoExecutionException:
Artifact has not been packaged yet. When used on reactor artifact, unpack should be executed after packaging.

This seems to happen during "copy-release-notes-from-core".within the depencency plugin.

PR #483, opened 2025-06-18
RAT-358: updated documentation for building RAT

Claudenw:
part of RAT-358


ottlinger:
As "clean install" is the default goal all JARs should be made available implicitly ....
should we just change the site-generation-helper scripts to run install instead of package?!

ottlinger:
@Claudenw I'm going to merge this as the build goes through locally in contrast to #484 - thanks for taking the time to document the problem.

PR #491, opened 2025-06-30
Rat 397 rebase

Claudenw:


Claudenw:
@ottlinger I rebased the RAT-397 feature branch onto main and this is the result.  It looks correct for all the changes outside of the site plugin.

For the site plugin it looks a bit weird.  Like it is not comparing against the proper branch -- though the result looks correct.

```

```

has been replaced with

```
<site xmlns="http://maven.apache.org/SITE/2.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/SITE/2.0.0 https://maven.apache.org/xsd/site-2.0.0.xsd" name="Apache Rat&#x2122; - A Release Audit Tool">
```

PR #521, opened 2025-09-12
Added information about RAT-476 to RAT-335 test

Claudenw:
* Adjusted RAT-355 test a per git reporting of include/exclude

Discovered RAT 335 was not 100% correct, then discovered that RAT-476 already covers the issue that I uncovered in RAT-335 test.  So rat 335 test is adjusted to show current state and has note to indicate why.

Also added comment to RAT-476 indicating that RAT-335 test should be updated when it is fixed.

### DIFF
--- a/apache-rat-core/src/it/resources/ReportTest/RAT_335/src/.gitignore
+++ b/apache-rat-core/src/it/resources/ReportTest/RAT_335/src/.gitignore
@@ -3,5 +3,6 @@
 # This makes it ignore dir3/dir3.log and dir3/file3.log
 *.log
 
-# This makes it "unignore" dir3/file3.log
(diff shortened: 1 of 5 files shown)


ottlinger (approved):
LGTM - thanks for linking the issues and verifying the behaviour with the integration test.

Claudenw:
> @Claudenw WDYT should we add a known issue section to the project homepage in order to report these known issues for any clients or add an info entry to the changelog about it?

If we add a known issues to the page we should have it link to the issue itself.  I think this makes sense.  Do you want to open a ticket for it and do we need to do it for RAT 17 release?

Claudenw:
I created https://issues.apache.org/jira/browse/RAT-489 to track the "Known Issues" page request and am merging this PR.

PR #525, opened 2025-09-14
Use the proper name for Apache License

Claudenw:
Differentiated between Apache Licenses 1.0, 1.1 and 2.0

Closes RAT-128


apache-rat-core/src/main/resources/org/apache/rat/asf.xml
```diff
@@ -33,9 +33,6 @@
 		<family id="UKOGL" name="UK Open Government License family" />
 		<family id="WTF" name="WTF Public License family" />
```
ottlinger, line 35: @Claudenw do you remember what this file was for as it enables GPL by default, which is not allowed in ASF projects?

ottlinger (approved):
Apart from the question related to asf.xml everything looks fine to me. 
Thanks for fixing this disambiguity! @Claudenw

apache-rat-core/src/main/resources/org/apache/rat/asf.xml
ottlinger: Should we remove the file? I did not find any references to it from the creadur-site repo. WDYT?

PR #542, opened 2025-10-04
RAT-475: Disable test on GHA windows/JDK8

ottlinger:
In order to stabilize release builds apart from LinuxGHA we could disable the dangling test on WindowsJDK8 as well.

WDYT?


Claudenw (approved):
LGTM.  Thx.

PR #546, opened 2025-10-06
Fix NPE on unsupported character set

Claudenw:
Change now redefines any document with an unsupported character set as an UNKNOWN document. Unable to create a test to verify.


Claudenw:
I think the deprecation report needs a bit more work.  Good catch, but the solution does not play well when only INFO logging is turned on.  I'll fix it.

PR #627, opened 2026-03-08
RAT-539: Update xml writer

Claudenw:
Update the XmlWriter to write to any appendable.
Move the XMLChar class to its own file and use the implementation for Xerces-j to pick up fixes.

Closes RAT-539


ottlinger (changes requested):
Thanks, LGTM - all sonarcloud warnings seem reasonable and should be applied before merging.

Besides I was just wonderung if more tests are needed to cover more of the corner cases that XmlWriter handles.

Thanks for all your work!

ottlinger (approved):

PR #637, opened 2026-03-22
RAT-540: Create UI option collections

Claudenw:
Introduces `ui` package 
Modifications to use `ui` package.

Changes of note:

- AbstractOption becomes UIOption and is UI Specific.
- Default values for parameters are moved into the UIOptionsCollection implementations.
- HELP option is removed from Arg and left as implementation detail in UIOptionsCollections as there is not a common implementation across the UIs.
- `ui` package is new and contains all the UI classes that will be used to create new UIs.  In future PRs the Ant and Maven UIs will be extracted and placed in separate modules.
- Many of the files are small changes to solve javadoc, checkstyle, or sonar issues that arise because of small changes to interfaces and  low level method signatures.

Fix for RAT-540


apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
```diff
@@ -337,7 +420,17 @@ public enum Arg {
                     .desc("Reads <Expression> entries from a file. Entries will be excluded from processing.")
                     .deprecated(DeprecatedAttributes.builder().setForRemoval(true).setSince("0.17")
                             .setDescription(StdMsgs.useMsg("--input-include-file")).get())
-                    .build())),
+                    .build()),
+            (context, selected) -> {
+                try {
+                    File includeFileName = context.getCommandLine().getParsedOptionValue(selected);
+                    if (includeFileName != null) {
+                        context.getConfiguration().addIncludedPatterns(ExclusionUtils.asIterable(includeFileName, "#"));
```
ottlinger, line 427: Description says "entries will be excluded from processing", but the method is "addIncludePatterns" - shouldn't it be "addExcludePatterns" then?

Most probably the description of the deprecated option is misleading ....

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
```diff
@@ -475,51 +653,66 @@ public enum Arg {
...
     private final OptionGroup group;
 
+    /**
+     * The BiConsumer to apply the option to update the state of the context.configuration.
+     */
+    private final BiConsumer<ArgumentContext, Option> process;
+
+
+    private static void doNotExecute(final ArgumentContext context, final Option selected) {
+        throw new ImplementationException(String.format("'%s' should not be executed directly", selected));
```
ottlinger, line 693: Initially I wondered "... what to do instead" is missing in that message to the user of RAT. 
If I get this message, did I do something wrong or not?

In the case of "help-licenses" I used the option correctly ....

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
```diff
@@ -806,82 +912,22 @@ public static void processLogLevel(final CommandLine commandLine) {
...
-            } else {
-                String[] style = context.getCommandLine().getOptionValues(OUTPUT_STYLE.getSelected());
-                if (style.length != 1) {
-                    DefaultLog.getInstance().error("Please specify a single stylesheet");
-                    throw new ConfigurationException("Please specify a single stylesheet");
-                }
-                context.getConfiguration().setStyleSheet(StyleSheets.getStyleSheet(style[0]));
-            }
+    private static void processOutputArgs(final ArgumentContext context, final UIOptionCollection<?> optionCollection) throws ConfigurationException {
+        for (Arg arg : new Arg[]{DRY_RUN, OUTPUT_FAMILIES, OUTPUT_LICENSES, OUTPUT_ARCHIVE, OUTPUT_STANDARD, OUTPUT_FILE, OUTPUT_STYLE}) {
```
ottlinger: Personally I'd prefer List.of() due to better readability than arrays here ....

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
```diff
@@ -692,91 +875,21 @@ private static void processConfigurationArgs(final ArgumentContext context) thro
...
-                } else {
-                    for (String s : context.getCommandLine().getOptionValues(selected)) {
-                        context.getConfiguration().addIncludedCollection(StandardCollection.valueOf(s));
-                    }
-                }
-            }
-        } catch (Exception e) {
-            throw ConfigurationException.from(e);
+    private static void processInputArgs(final ArgumentContext context, final UIOptionCollection<?> optionCollection) throws ConfigurationException {
+        for (Arg arg : new Arg[]{SOURCE, EXCLUDE, EXCLUDE_FILE, EXCLUDE_STD, EXCLUDE_PARSE_SCM, EXCLUDE_SIZE,
```
ottlinger: Personally I'd prefer List.of() due to better readability than arrays here ....

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
```diff
@@ -625,64 +844,28 @@ private static List<String> processArrayFile(final ArgumentContext context, fina
...
+                    }
+                });
+        optionCollection.getSelected(CONFIGURATION_NO_DEFAULTS).ifPresent(selected -> {
+            // display deprecation log if needed.
+            context.getCommandLine().hasOption(selected);
+            defaultBuilder.noDefault();
+        });
+        context.getConfiguration().setFrom(defaultBuilder.build());
+
+        for (Arg arg : new Arg[]{FAMILIES_APPROVED, FAMILIES_APPROVED_FILE, FAMILIES_DENIED, FAMILIES_DENIED_FILE,
```
ottlinger: Personally I'd prefer List.of() due to better readability than arrays here ....

apache-rat-core/src/main/java/org/apache/rat/ui/spi/UIProvider.java
```diff
@@ -0,0 +1,26 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.rat.ui.spi;
+
+import org.apache.rat.ui.UI;
+import org.apache.rat.ui.UIOption;
+
```
ottlinger, line 23: Should we add a general description here that explains the proposed architecture?

Claudenw:
@ottlinger, Once you finish this review I will rebase onto main to clean up the build errors.

apache-rat-core/src/test/java/org/apache/rat/ReporterOptionsProvider.java
```diff
@@ -1275,7 +1275,6 @@ protected void editOverwriteTest() {
     }
 
     @OptionCollectionTest.TestFunction
-    @Override
```
ottlinger, line 1278: The class contains many unused (reported by IDE) methods - is that part of the changes or can these methods be removed?

apache-rat-plugin/src/it/RAT-508/pom.xml
```diff
@@ -34,12 +34,8 @@
...
           <consoleOutput>true</consoleOutput>
           <ignoreErrors>false</ignoreErrors>
-          <inputIncludes>
-            <include>pom.xml</include>
-          </inputIncludes>
-          <inputExcludeStds>
-            <exclude>ALL</exclude>
-          </inputExcludeStds>
+          <inputIncludes>pom.xml</inputIncludes>
+          <inputExcludeStds>ALL</inputExcludeStds>
```
ottlinger, line 38: Should we document this breaking change in the change-notes for 1.0.0 explicitly?

apache-rat-plugin/src/test/resources/unit/it5/pom.xml
```diff
@@ -28,12 +28,12 @@
...
-          <config>.rat/customConfig.xml</config>
-          <inputExcludes>
-            <exclude>.rat/**</exclude>
-            <exclude>pom.xml</exclude>
-            <exclude>invoker.properties</exclude>
-          </inputExcludes>
+          <configs>
+            <config>.rat/customConfig.xml</config>
+          </configs>
+          <inputExcludes>.rat/**</inputExcludes>
```
ottlinger, line 34: I'm not sure if this configuration is "Maven-style" - would it make sense to ask over at the Maven mailing list?

plugins contains multiple plugin-entries .... that made me think about this breaking change.

apache-rat-tools/src/main/java/org/apache/rat/tools/AntDocumentation.java
```diff
@@ -79,13 +77,13 @@ public static void main(final String[] args) {
         new AntDocumentation(outputDir).execute();
     }
 
-   private AntDocumentation(final File outputDir) {
+    /* Visible for testing */
```
ottlinger, line 80: Class contains unused methods:
* writeTitle
* writePara
* writeList

Should they be removed or will they be needed later and are only yet unused?

ottlinger (changes requested):

apache-rat-plugin/src/test/resources/unit/it5/pom.xml
ottlinger: What I mean is: how does a configuration with multiple entries look like?

```
          <inputIncludes>
            <include>pom.xml</include>
            <include>foo.bar/include>
          </inputIncludes>

```
OR will it become
```
          <inputIncludes>pom.xml</inputIncludes>
          <inputIncludes>foo.bar</inputIncludes>

```

apache-rat-tools/src/main/java/org/apache/rat/tools/AntDocumentation.java
Claudenw: As I recall, this class and methods will be used in the Ant UI package.

I wouldn't worry too much about tools as it goes away at the end of this trasformation.

In the end we will have the following modules

- core: (the core functionality)
- core test:  test code from core packaged with dependencies to make it easier for UI modules.  Basically a maven module that points to core `/src/test` for its `/src/main`.  Jars built like this include dependencies so that when included as test libraries the transitive dependencies are present.
- CLI: the current command line interface
- ANT: the ANT interface and components from tools that are used for it.
- Maven: the Maven interface and the components from tools that are used for it.
- packaging: -- Not sure about this one, but I think that it just puts outputs from the above modules into release (as in upload to sonatype or github release style) packages and deploys them.

The UI modules CLI, ANT, and Maven will have multiple sub modules.  Something like

UI module pom
  - tools:  Tooling from current tool module + additional bits.
  - main: The actual code that is released.  Module should build the packaging necessary to use the UI as a stand alone unit

Claudenw:
@ottlinger you have a number of good questions and ideas here with respect to how this works going forward.  I like the idea of describing how the architecture works.  Would you rather I fix/comment on the issues in this PR (and potentially make it larger) or open another PR to add the documentation.

I had planned on writing/updating the general documentation (.md files) when this change is complete.

ottlinger:
@Claudenw I created RAT-557 to add documentation and have a new PR for it. The Maven/configuration stuff should be handled here I guess.

apache-rat-plugin/src/it/RAT-508/pom.xml
ottlinger: Created RAT-556 as a reminder ticket

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
Claudenw: good catch.

apache-rat-core/src/test/java/org/apache/rat/ReporterOptionsProvider.java
Claudenw: The class is not used yet.  Next step is to start using these for Ant or Maven.

apache-rat-plugin/src/it/RAT-508/pom.xml
Claudenw: Thx

apache-rat-plugin/src/test/resources/unit/it5/pom.xml
Claudenw: I asked on the Maven dev mailing list for guidance on the standard Maven way.  We will follow whatever comes from that.

Claudenw:
@ottlinger I made the `List.of` substitution for `new Array[]` as you suggested.  If you are happy with the result please merge.

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
ottlinger: Is there change needed here? WDYT?

ottlinger (changes requested):
Is this includeExclude thing a bug?

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
Claudenw: I changed it to read:

```
INCLUDE_FILE(new OptionGroup()
            .addOption(Option.builder().longOpt("input-include-file")
                    .argName("File").hasArg().type(File.class)
                    .converter(Converters.FILE_CONVERTER)
                    .desc("Reads <Expression> entries from a file. Entries will override excluded files.")
                    .build())
            .addOption(Option.builder().longOpt("includes-file")
                    .argName("File").hasArg().type(File.class)
                    .converter(Converters.FILE_CONVERTER)
                    .desc("Reads <Expression> entries from a file. Entries will override excluded files.")
                    .deprecated(DeprecatedAttributes.builder().setForRemoval(true).setSince("0.17")
                            .setDescription(StdMsgs.useMsg("--input-include-file")).get())
                    .build()),
            (context, selected) -> {
                try {
                    File includeFileName = context.getCommandLine().getParsedOptionValue(selected);
                    if (includeFileName != null) {
                        context.getConfiguration().addIncludedPatterns(ExclusionUtils.asIterable(includeFileName, "#"));
                    }
                } catch (Exception e) {
                    throw ConfigurationException.from(e);
                }
            }),
```

So I think the documentation will be correct.

apache-rat-plugin/src/test/resources/unit/it5/pom.xml
Claudenw: OK.  I have an answer and will implement the changes when I split the maven UI into its new module

apache-rat-core/src/main/java/org/apache/rat/commandline/Arg.java
Claudenw: Added javadoc to explain what this method does.

apache-rat-plugin/src/test/resources/unit/it5/pom.xml
Claudenw: I updated https://issues.apache.org/jira/browse/RAT-543 with a link to the discussion about how to implement.

Claudenw:
> Is this includeExclude thing a bug?

I don't think so.   The rule is:

1. Everything is included by default
2. things can be explicitly excluded to override default inclusion.
3. things can be explicitly include to override explicit exclusion.  Explicitly included things may not be excluded again.

I think the documentation for the option was a bit off.

PR #653, opened 2026-04-27
RAT-551: Replace NoCloseOutputStream with Apache Commons CloseShieldOutputStream

Claudenw:
Removes custom NoCloseOutputStream with the Apache commons CloseShieldOutputStream


ottlinger:
@Claudenw is it merely to use existing library code or is there a different motivation behinde the change?

Sonar fails due to not knowing the context of the code, but can be ignored.

ottlinger (approved):
LGTM - do we have a code that explicitly triggers this or is it done implicitly by integration/example docs?

Claudenw:
NoCloseOutputStream was custom rolled and caused problems with some checkers as they could not determine if the OutputStream was closed.  The Apache version works well with checkers and is included in a library we already use.

So simplifies our code.
Makes code cleaner for checkers.

PR #658, opened 2026-05-11
RAT-502: Map XSLT to standard instead of binary files 

ottlinger:
Scan XSLT fopr licenses instead of ignore them as binary.
Fixed regression from 0.17 and fixed some typos as well.


Claudenw (approved):
Looks good to me.  Thx

PR #662, opened 2026-05-19
RAT-555: Configure TempDir deletion strategy to ignore I/O failures and skip running on GHA&linux

ottlinger:
Random build failures seem to be back on GHA, e.g.
https://github.com/apache/creadur-rat/pull/660

Try out freshly implemented DeletionStrategy in jUnit's TempDir and exclude running the test on (GHA&&linux)

### DIFF
--- a/apache-rat-tasks/src/test/java/org/apache/rat/anttasks/ReportOptionTest.java
+++ b/apache-rat-tasks/src/test/java/org/apache/rat/anttasks/ReportOptionTest.java
@@ -37,6 +37,7 @@ import org.apache.rat.utils.Log;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.io.CleanupMode;
 import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.api.io.TempDirDeletionStrategy;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ArgumentsProvider;
 import org.junit.jupiter.params.provider.ArgumentsSource;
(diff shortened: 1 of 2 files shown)


ottlinger:
@Claudenw can you review and merge? thx

apache-rat-tasks/src/test/java/org/apache/rat/anttasks/ReportOptionTest.java
```diff
@@ -53,9 +55,11 @@
 /**
  * Tests to ensure the option setting works correctly.
  */
+@DisabledIf(value = "isRunningOnGitHubActionOrLinux", disabledReason = "RAT-555, RAT-475")
```
Claudenw: `or linux`?  Should this not be `and linux`? or `with linux` so it is clear that it is only disabled on github running linux.    

Also, I think the implementation of `isRunningOnGitHubActionOrLinux`  is incorrect. as it should only return true with linux on github.

Claudenw (changes requested):

apache-rat-tasks/src/test/java/org/apache/rat/anttasks/ReportOptionTest.java
ottlinger: On ASF Jenkins we had test failures, but I will adapt the implementation. Thanks.

ottlinger: Changed method and cleaned up code. Thx

ottlinger:
@Claudenw can we merge now WDYT?

Claudenw (approved):
Thanks

PR #671, opened 2026-06-01
RAT-558: Add proposals for AGENTS and SECURITY configuration - to be discussed

ottlinger:
This PR is supposed to start a discussion concercing agent and security configuration of RAT.

Other examples within the ASF are:
* https://github.com/apache/grails-core/

PR #677, opened 2026-06-10
RAT-558: Add security threat model (THREAT_MODEL.md + SECURITY.md + AGENTS.md)

potiuk:
## What

Adds a **threat model** for Apache Creadur (RAT) at the Creadur PMC's request (GLASSWING / Mythos scan pre-flight):

- **`THREAT_MODEL.md`** — the model ([rubric](https://gist.github.com/potiuk/da14a826283038ddfe38cc9fe6310573)).
- **`SECURITY.md`** + **`AGENTS.md`** — disclosure pointer + the `AGENTS.md -> SECURITY.md -> THREAT_MODEL.md` chain.

## The model in brief

RAT is modelled as an **in-process build/CLI license-audit tool** — not a network service, and explicitly **not a security/vulnerability scanner**. Its security-relevant case is auditing **untrusted** input: the XML configuration (XXE surface) and archive descent (decompression-bomb surface). Findings that require RAT to process input the operator already trusts (the normal case — your own source tree) are out of model.

## DRAFT — you own it; two quick technical confirmations

Because RAT is small, the §8-vs-§9 split hinges on two facts I've left as **section 14 questions**:
- **Q3** — does `XMLConfigurationReader` disable DOCTYPE/external entities (XXE-safe)?
- **Q4** — does `ArchiveWalker` bound decompression (size/depth/entry-count)?

Your answers turn those from "open question" into either a provided property (§8) or a documented gap + downstream note (§9). Also **Q6**: want me to add the same chain to `creadur-whisker` and `creadur-tentacles` so all three are discoverable?

Generated by the ASF Security team's threat-model tooling (Claude Opus); reviewed before opening.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,3 +31,10 @@ consult before producing output.
 - Don't add unnecessary dependencies
 - Follow the existing codebase patterns and conventions
 - Test your solutions when possible with unit or integration tests
+
+## Security
+
+Security model: [SECURITY.md](./SECURITY.md), which links to the project's
+threat model at [THREAT_MODEL.md](./THREAT_MODEL.md). Consult the threat model
+for the project's in-scope / out-of-scope declarations and known non-findings
+before reporting security issues.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,40 +1,17 @@
-# Agent guidance
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
 
-This file is read by automated agents (security scanners, code
-analyzers, AI assistants) operating on this repository. It
-points them at the human-authored references they should
-consult before producing output.
+      https://www.apache.org/licenses/LICENSE-2.0
 
-# CODE STYLE
-- Use consistent naming conventions
-- Follow language-specific style guides
-- Keep functions small and focused
-- Use meaningful variable and function names
-- Add comments for complex logic
-- Add package-info.java to each new package in the source tree
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
 
-# BEST PRACTICES
-- **DRY (Don't Repeat Yourself)**: Avoid code duplication
-- **SOLID Principles**: Follow object-oriented design principles
-- **Error Handling**: Always handle potential errors gracefully
-- **Security**: Consider security implications in your code
-- **Version Control**: Write clear commit messages, referencing Apache Creadur Jira tickets if possible
-
-# COMMUNICATION
-- Explain your approach before implementing
-- Break down complex solutions into steps
-- Provide examples when helpful
-- Ask clarifying questions when requirements are unclear
-
-# RESTRICTIONS
-- Always ask before making breaking changes
-- Don't add unnecessary dependencies
-- Follow the existing codebase patterns and conventions
-- Test your solutions when possible with unit or integration tests
-
-## Security
-
-Security model: [SECURITY.md](./SECURITY.md), which links to the project's
-threat model at [THREAT_MODEL.md](./THREAT_MODEL.md). Consult the threat model
-for the project's in-scope / out-of-scope declarations and known non-findings
-before reporting security issues.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,4 +14,45 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
+# Agent guidance
 
+This file is read by automated agents (security scanners, code
+analyzers, AI assistants) operating on this repository. It
+points them at the human-authored references they should
+consult before producing output.
+
+# SECURITY
+- Before reporting any security finding, consult [`SECURITY.md`](SECURITY.md)
+  and the full threat model in [`THREAT_MODEL.md`](THREAT_MODEL.md).
+- `THREAT_MODEL.md` defines RAT's scope, trust boundaries, the security
+  properties RAT provides and disclaims, the adversary model, and the
+  known non-findings used to triage scanner output.
+- Report suspected vulnerabilities privately to security@apache.org; do not
+  open public issues for them.
+
+# CODE STYLE
+- Use consistent naming conventions
+- Follow language-specific style guides
+- Keep functions small and focused
+- Use meaningful variable and function names
+- Add comments for complex logic
+- Add package-info.java to each new package in the source tree
+
+# BEST PRACTICES
+- **DRY (Don't Repeat Yourself)**: Avoid code duplication
+- **SOLID Principles**: Follow object-oriented design principles
+- **Error Handling**: Always handle potential errors gracefully
+- **Security**: Consider security implications in your code
+- **Version Control**: Write clear commit messages, referencing Apache Creadur Jira tickets if possible
+
+# COMMUNICATION
+- Explain your approach before implementing
+- Break down complex solutions into steps
+- Provide examples when helpful
+- Ask clarifying questions when requirements are unclear
+
+# RESTRICTIONS
+- Always ask before making breaking changes
+- Don't add unnecessary dependencies
+- Follow the existing codebase patterns and conventions
+- Test your solutions when possible with unit or integration tests
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -30,11 +30,11 @@ not report security issues on public issue trackers or mailing lists.
 
 ## Known Non-Findings
 
-- Static code analysis may report `XXE_DOCUMENT` vulnerabilities because RAT reads XML and XSLT files provided as user input.
+- Static analyzers may report `XXE_DOCUMENT` on RAT's XML/XSLT reading. As of RAT-560 ([#679](https://github.com/apache/creadur-rat/pull/679)) RAT builds its XML parsers through the hardened `StandardXmlFactory`, which disables DOCTYPE and external general/parameter entities — so XXE is actively prevented and these reports are false positives against the hardened factory.
 
-  - Configuration files and XSLT documents passed to RAT are operator-controlled configuration, not request input. Reports claiming SSRF or path traversal via these resolvers, based on the assumption that the resource name is attacker-controlled, are out of scope under the documented threat model. XML and XSLT authorship, as well as resource configuration, are privileged operations.
+  - Defense in depth: the configuration files and XSLT documents RAT reads are operator-controlled configuration, not request input, so the resource names are not attacker-controlled in the first place. Reports asserting SSRF or path traversal via these resolvers (assuming an attacker-controlled resource name) are out of scope under the documented threat model — XML and XSLT authorship, as well as resource configuration, are privileged operations.
 
-  - Applications that thread untrusted input into XML configuration or XSLT documents should validate that input before passing it to RAT. Responsibility for such validation rests with the application, not with RAT.
+  - Applications that thread untrusted input into XML configuration or XSLT documents should still validate that input before passing it to RAT. Responsibility for such validation rests with the application, not with RAT.
 
 ## Threat Model
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -9,3 +9,11 @@ TBD
   - Configuration files and XSLT documents passed to RAT are operator-controlled configuration, not request input. Reports claiming SSRF or path traversal via these resolvers, based on the assumption that the resource name is attacker-controlled, are out of scope under the documented threat model. XML and XSLT authorship, as well as resource configuration, are privileged operations.
 
   - Applications that thread untrusted input into XML configuration or XSLT documents should validate that input before passing it to RAT. Responsibility for such validation rests with the application, not with RAT.
+
+## Threat Model
+
+The full Apache Creadur RAT threat model — scope and intended use, trust
+boundaries, the security properties RAT provides and disclaims, the adversary
+model, and known non-findings — is documented in
+[THREAT_MODEL.md](./THREAT_MODEL.md). The scope notes above are a summary;
+THREAT_MODEL.md is the detailed companion.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,6 +1,32 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
 # Security
 
-TBD
+Apache Creadur RAT (Release Audit Tool) runs as a CLI, an Ant task, or a Maven
+plugin in the developer's or CI's own process — it is not a network service. It
+audits a source tree against operator-controlled license and header definitions.
+
+## Reporting a Vulnerability
+
+Please report suspected security vulnerabilities privately to the Apache Security
+Team at security@apache.org, following the
+[ASF vulnerability handling process](https://www.apache.org/security/). Please do
+not report security issues on public issue trackers or mailing lists.
 
 ## Known Non-Findings
 
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -1,0 +1,247 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+# Apache Creadur (RAT) — Threat Model
+
+## §1 Header
+
+- **Project:** Apache Creadur — primarily **RAT (Release Audit Tool)**
+  (`apache/creadur-rat`), with sibling tools **Whisker**
+  (`apache/creadur-whisker`, license-documentation generator) and **Tentacles**
+  (`apache/creadur-tentacles`, release-bundle analyzer). This model is written
+  in `creadur-rat` and covers the Creadur dev-tool family; Whisker/Tentacles
+  share RAT's trust profile (§2).
+- **Written against:** `main`/`master` @ HEAD (2026-06).
+- **Author:** ASF Security team, via the threat-model-producer rubric (Scovetta
+  rubric) at the Creadur PMC's request (path 3).
+- **Status:** DRAFT — under maintainer review (2026-06-10). Not yet ratified.
+- **Reporting cross-reference:** §8-violating findings via the ASF security
+  process ([`SECURITY.md`](SECURITY.md)); §3/§9 findings closed citing this doc.
+- **Provenance legend:** *(documented)* / *(maintainer)* / *(inferred)* — each
+  *(inferred)* has a §14 open question.
+- **Draft confidence:** ~14 documented / 0 maintainer / 16 inferred.
+
+**What it is.** RAT is a **build-time / CLI license-auditing tool**: it walks a
+source tree, matches files against configurable license/header definitions, and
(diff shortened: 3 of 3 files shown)


Claudenw:
This PR looks like it needs answers from developers before submitting.
@potiuk is that the expected flow?  We answer questions and push the resulting file(s) to the repo?

potiuk:
Yes. Absolutely. - it's enough we just comment in the PR answering the questions and I will update the PR accordingly

THREAT_MODEL.md
```diff
@@ -0,0 +1,247 @@
...
+  *(inferred)* has a §14 open question.
+- **Draft confidence:** ~14 documented / 0 maintainer / 16 inferred.
+
+**What it is.** RAT is a **build-time / CLI license-auditing tool**: it walks a
+source tree, matches files against configurable license/header definitions, and
+reports unapproved or unknown licenses. It runs as a **CLI**, an **Ant task**,
+or a **Maven plugin** — always **in the developer's or CI's own process**,
+never as a network service. Whisker generates license documentation; Tentacles
+inspects staged release bundles. None is a server.
+
```
ottlinger, line 45: Would it make sense to add a new notion here as RAT can be used to change your own sources to include license headers? In this way user input can be altered or is this not relevant from a security-scope?

THREAT_MODEL.md
```diff
@@ -0,0 +1,247 @@
...
+  a vulnerability scanner or a security gate. "RAT didn't catch X security
+  issue" is not in scope. *(documented — purpose.)*
+- **Audit *correctness* as a security property.** A missed/false license match
+  is a correctness bug, not a vulnerability (unless it crosses a resource bound,
+  §8). *(inferred.)*
+- **The build/CI environment** RAT runs in, and the trust of the source tree
+  when RAT is deliberately run on your own (trusted) code — the dominant,
+  intended case. Findings whose only impact requires running RAT on input you
+  already trust are `OUT-OF-MODEL: trusted-input`.
+- **Test resources** (the deliberately-odd license fixtures under
```
ottlinger, line 88: Would it make sense to add documentation that Maven/Ant/CLI options is generated, as this would mean that security implications are automatically transferred to all of RAT UIs?

THREAT_MODEL.md
```diff
@@ -0,0 +1,247 @@
...
+**Wave 3 — coexistence.**
+
+- **Q6.** This adds `THREAT_MODEL.md` + `SECURITY.md` + `AGENTS.md` to
+  `creadur-rat`. Want matching pointer files (AGENTS.md → SECURITY.md → this
+  model) added to `creadur-whisker` and `creadur-tentacles` so all three are
+  discoverable, or will you add them? (§1/§15.)
+
+## §15 Appendix — existing-policy back-map
+
+No in-repo `SECURITY.md` exists today; this PR adds one (ASF security-process
```
ottlinger: a very basic security file was introduced via 
https://github.com/apache/creadur-rat/pull/671

Claudenw:
@potiuk 

In answer to the first question, there is a PR to ensure we have this  covered.
In answerto the second question, we utilize Apache Common Compress ArchiveStreamFactory class to create archive streams.  If the filters indicate that we should process the archived file we extract the contents into an in-memory buffer where it is held until the document is processed.  So we could hit OOM.

potiuk:
Thanks `Claudenw` and `ottlinger` — folded your review into the draft and pushed (THREAT_MODEL.md, +53/-21):

- **Archive handling (your Q on OOM):** confirmed as a gap. The unbounded in-memory extraction via Commons Compress `ArchiveStreamFactory` is now §9 (not a provided property), with the matching §10 downstream responsibility (sandbox / resource-limit RAT runs over untrusted archives) and a VALID-HARDENING disposition. Marked maintainer-confirmed.

- **XML/DOCTYPE (your "there is a PR to ensure we have this covered"):** I read that as the XXE-hardening question (§14 Q3). I've noted a hardening PR is in flight but left §8 #2 tentative pending the link — could you drop the PR number here so I can cite it? Once it lands I'll flip XXE to a provided property.

- **`ottlinger` — write mode:** good catch. RAT's `--addLicense` / editor mode *writes* headers into the audited tree; I've documented it explicitly in §2/§3 as operator-invoked on the operator's own (trusted) sources → it's OUT-OF-MODEL: trusted-input, but now stated rather than silent.

- **`ottlinger` — generated front-ends:** added a §2 note that CLI/Ant/Maven are generated from a common option core, so a security property (or gap) in the core transfers to all three UIs.

- **`ottlinger` — SECURITY.md / #671:** thanks, fixed the stale §15 line — it now reflects that #671 added SECURITY.md and this PR just appends the AGENTS.md -> SECURITY.md -> THREAT_MODEL.md pointer.

Still open if you have a moment (one line each is plenty): Q1 (confirm the untrusted-input case is the one to model), Q2 (RAT makes no network connections), Q5 (Whisker/Tentacles share the profile), and Q6 (want us to add the same pointer files to creadur-whisker/-tentacles, or will you?).

One note on CI: the failing "Build and analyze" (CodeQL) check is unrelated to this PR — it's a docs-only change (three .md files), so it isn't introducing or affected by that build job; looks pre-existing/flaky on the branch.

Claudenw:
@potiuk There is one more point that has not been discussed.  RAT allows developers to extend the matching algorithms.  See https://creadur.apache.org/rat/license_def.html#Matchers

The upshot is that 3rd parties can create new matchers and use them in license checks.  Matchers are different from license checks in that license checks use matchers.  For example the Apache 2.0 license check uses the `text`, `spdx` and `any` matchers.

Matchers scan the contents of the file (as a String) looking for matches.  This means that a custom matcher would have access to all text from all files that are selected for scanning.

But this is defined in the configuration and is under control of the developer using RAT.

ottlinger:
@potiuk the Sonarbuild does only run on specific branches/with specific PRs as the credentials are not shared among all PRs/builds due to ASF restrictions.

Claudenw:
> * could you drop the PR number here so I can cite it? Once it lands I'll flip XXE to a provided property.

https://github.com/apache/creadur-rat/pull/679 is the PR that does the XXE hardening.

potiuk:
Thanks `Claudenw` — folded the custom-matcher surface into §3: RAT lets operators define custom matcher classes that see the full text of every scanned file, but since the matcher set is operator-defined config (not attacker-supplied), a custom matcher reading scanned text is `OUT-OF-MODEL: trusted-input` — the same posture as the write mode. Marked maintainer-confirmed.

`ottlinger` — thanks for the CI clarification; that matches what we expected (the Sonar/CodeQL job is credential-gated to specific branches under ASF restrictions, so its red on this fork PR is environmental, not anything this docs-only change introduces).

Still one open item: the XXE-hardening PR number (§14 Q3) — I've left §8 #2 tentative pending it. Whenever you drop the number I'll cite it and flip XXE from "hardening in flight" to a provided property. No rush.

The remaining §14 questions (Q1 untrusted-input posture, Q2 no-network, Q5 Whisker/Tentacles profile, Q6 sibling pointer files) are still open whenever convenient — one line each is plenty.

ottlinger:
@potiuk - thanks again:

* Q5:as development on Tentacles/Whisker is rather low at the moment I'd personally prefer to start with RAT and add Tentacles/Whisker later in order to reduce noise.

* Q2: AFAIK RAT does not open network connections as it runs locally; except for downloading stuff via Maven/Ant/buildtools in order to build the artifacts locally.

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+license compliance before a release or on each change. The two inputs are the
+**tree being audited** (files, including archives RAT descends into) and the
+**RAT configuration** (XML/text license + matcher definitions).
+
+Caller trust level: the developer/CI invoking RAT is trusted. The **inputs are
+normally trusted too** (your own source, your own config) — but RAT is
+sometimes pointed at **untrusted input**: a CI job auditing an untrusted
+contribution/PR, or auditing a downloaded third-party artifact. That is the
+case the model cares about. *(inferred — Q1.)*
+
```
Claudenw, line 59: This statement is correct

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+## §4 Trust boundaries and data flow
+
+The boundary is **the input RAT is pointed at** — files and configuration.
+RAT's security questions only arise when that input is **untrusted**:
+
+```
+caller invokes RAT (CLI/Ant/Maven) on a directory + a config
+   │ trusted invocation
+   ▼
+read configuration (XMLConfigurationReader) ── XXE surface if config is untrusted
```
Claudenw: https://github.com/apache/creadur-rat/pull/679 is the PR that does the XXE hardening.  I don't know if that impacts here.

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+is triggered by **untrusted input** (a hostile file/archive/config) that a
+*realistic* RAT deployment processes — e.g. CI auditing an untrusted PR. A
+finding that requires the operator to feed RAT input they already control is
+`OUT-OF-MODEL: trusted-input` (§3).
+
+## §5 Assumptions about the environment
+
+- A JRE; RAT reads the filesystem it is pointed at and writes a report. It opens
+  **no network connections** and runs no services. *(inferred — Q2, the
+  no-network claim is high-value to confirm.)*
```
Claudenw: True, no network connections are opened by RAT.  RAT only opens files.  One potential hole in this is XSLT transforms where the operator could add an xsl:include statement to open a connection to a remote system.  This is out of scope as the XSLT are in the trusted space under control of the operator.

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+- The XML parser behaviour depends on the platform JAXP unless RAT configures it
+  (§5a/§8). *(inferred — Q3.)*
```
Claudenw: It depends upon JAXP an can be configured through the JAXP environment variables as documented:
https://docs.oracle.com/javase/8/docs/technotes/guides/security/jaxp/jaxp.html#setting-jaxp-properties-as-system-properties

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+RAT has no security-mode flag. The security-relevant configuration is whether
+its **XML config parser disables DOCTYPE/external entities** and whether the
+**archive walker bounds decompression** (depth/size/entry count). Both are
+hardcoded behaviours, not operator knobs. The **archive walker does not bound
+decompression** — it extracts entry contents into an in-memory buffer (Apache
+Commons Compress `ArchiveStreamFactory`) with no size/depth/entry-count limit
+(§8/§9, maintainer-confirmed). XML-parser DOCTYPE handling is being hardened via
+a PMC PR (§14 Q3). There is no "insecure default toggle". *(maintainer / Q3
+pending PR link.)*
+
```
Claudenw, line 153: See above for PR link

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+## §6 Assumptions about inputs
+
+| Input | Attacker-controllable? (untrusted-run) | Concern |
+| --- | --- | --- |
+| scanned file content | **yes** | parsed/read; resource use |
+| scanned file paths / archive entry names | **yes** | path handling on archive extraction |
```
Claudenw: What does "path handling on archive extraction" mean?  We do not extract the data into a directory.  We read the files from the archive and extract them from there.  The file paths are documented as relative to the archive so something like "/bar/baz.zip#/junk.txt" is reported for a file `junk.txt` in the archive baz.zip found in the `/some/dir/bar/` directory on a unix/mac system where RAT was pointed to `/some/dir` as the tree to scan.

But the contents of `junk.txt` was only every extracted to memory.

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+
+## §14 Open questions for the maintainers
+
+**Wave 1 — the load-bearing ones.**
+
+- **Q1.** Confirm the intended trust posture: RAT runs in-process for a
+  trusted caller; inputs are normally trusted, but the security-relevant case is
+  RAT auditing **untrusted** input (CI on untrusted PRs, third-party artifacts).
+  Is that the case you want modelled, or do you consider all RAT input trusted
+  (which would move XXE/archive items to `OUT-OF-MODEL: trusted-input`)? (§2/§7.)
```
Claudenw: All rat configuration items (XSLT stylesheets, configuration files, license definitions, matcher implementations) are trusted and under control of the operator.  

The files that are read may be untrusted, as you point out, in the case of verification of PRs from 3rd parties.  

Attack surface is anything that can break out of the scanning stream when the system is run with the default settings.  I am certain that there are settings that could open the system up for attack, for example the JAXP environment vars.s

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+- **Q3.** *(Partially answered — PMC, PR #677: a hardening PR is in flight
+  ensuring DOCTYPE / external-entity handling is covered. **Pending the PR link
+  to cite**; once landed §8 #2 becomes a provided property.)* Does
+  `XMLConfigurationReader` disable DOCTYPE / external entities (XXE-safe)?
```
Claudenw: External entities are disabled.

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+- **Q4.** *(Answered — PMC, PR #677: no bound. Archives are extracted into an
+  in-memory buffer (Commons Compress `ArchiveStreamFactory`) held until the
+  document is processed, so a crafted archive can OOM. Resolved as a §9 gap +
+  §10 responsibility; §8 #1 is **not** a provided property.)* Does
+  `ArchiveWalker` bound decompression (size/depth/entry-count)?
```
Claudenw: We do not guard against OOM, we probably should add a limit but at this time we do not.  This is out of scope and not provided.

THREAT_MODEL.md
```diff
@@ -0,0 +1,288 @@
...
+
+**Wave 2 — surface.**
+
+- **Q2.** Confirm RAT makes no network connections and has no side effects beyond
+  reading the scanned tree and writing the report. (§5/§8.)
```
Claudenw: Confirmed above.

potiuk:
Thanks @ottlinger and @Claudenw — that's everything we needed. I've folded all your answers into `THREAT_MODEL.md` (pushed). Summary of what changed:

- **Trust posture (Q1)** — confirmed and now `(maintainer)`: RAT config (XSLT, config files, license defs, custom matchers) is operator-trusted; the *scanned files* are the untrusted surface. §2/§7.
- **No network (Q2)** — confirmed `(maintainer)`. Added your XSLT `xsl:include` nuance: the one operator-reachable way out is a remote `xsl:include`, and since stylesheets are trusted config that's `OUT-OF-MODEL`. §5/§8 #3.
- **XXE (Q3)** — external entities are **disabled**, so §8 #2 is now a **provided** property (was tentative); noted PR #679 as the DOCTYPE-hardening follow-up, and the JAXP-system-properties configurability. §5/§5a/§8.
- **Archive bound (Q4)** — kept as a disclaimed §9 gap (no bound, OOM not guarded).
- **Path handling** — corrected a phantom risk: since RAT reads entries into memory and **never extracts to disk**, there's no zip-slip / path-traversal-on-write surface. An entry label like `bar/baz.zip#/junk.txt` is just a report string. §6/§9.
- **Whisker/Tentacles (Q5/Q6)** — scoped this PR to `creadur-rat` per your preference; the sibling pointer files are a deferred follow-up.

With every §14 question answered, the model is ready to ratify whenever the PMC's happy with it. (The red check is the CodeQL "Build and analyze" job, which is unrelated to these doc-only changes — all 13 build/test matrix jobs pass.)

ottlinger:
@potiuk if we merge this PR we are done and trigger the next step or are there any other things that need to be done on our side? Thanks

potiuk:
@ottlinger @Claudenw — quick note on the red check: that's only the SonarCloud job, which can't run on PRs because the ASF doesn't share its Sonar credentials with PR builds (as you noted). It's not an actual test failure, so it's safe to merge past it. And yes — merging this is the last step on your side; once it lands I'll verify the AGENTS.md → SECURITY.md → THREAT_MODEL.md chain resolves and Creadur/RAT enters the queue. (Starting with RAT and adding Tentacles/Whisker later is fine.)

ottlinger:
@Claudenw anything to add from your side? I'm fine with merging in order to go forward. Thx

ottlinger (approved):
LGTM as a first draft

THREAT_MODEL.md
```diff
@@ -0,0 +1,302 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
```
Claudenw: can we change this to https:

AGENTS.md
Claudenw, line 17: There is no license header on this file.

SECURITY.md
Claudenw, line 18: No license header on this file.

Claudenw: Have we determined what goes in here yet?

THREAT_MODEL.md
Claudenw: Rat can change the source that it is scanning, but it uses the text from a controlled input file.

THREAT_MODEL.md
Claudenw: Why does this still mention the XXE surface?

Claudenw (changes requested):
I will have to spend more time thinking about the threat model.  I think with license headers, and the XXE questions resolved we can proceed.

THREAT_MODEL.md
potiuk: Thanks both — and Claude, your "controlled input file" point nails it. I've folded it into the §3 write-mode note: it now spells out that the header text `--addLicense` writes in comes from the operator-controlled license/header definition (that controlled input file), not from the scanned tree, so write mode can't inject attacker-chosen content. (a6978b8)

THREAT_MODEL.md
potiuk: Good thinking — and it turns out this is already captured a little further down: the §2 "Note (PMC, review)" spells out that the CLI / Ant / Maven front-ends are generated from a common option core, so any security-relevant behaviour (or gap) transfers automatically to all three UIs. Resolving against that — thanks!

THREAT_MODEL.md
potiuk: Done — switched to https. (a6978b8) Thanks Claude!

AGENTS.md
potiuk: Good catch — ASF license header added to AGENTS.md. (a6978b8) Thanks!

SECURITY.md
potiuk: Fixed here too — license header added to SECURITY.md. (a6978b8) Thanks Claude!

potiuk: Filled it in (a6978b8): replaced the TBD with a short security-posture summary (RAT runs as a CLI / Ant task / Maven plugin in your own process, not a network service) plus a "Reporting a Vulnerability" section pointing at security@apache.org. The detailed model stays in THREAT_MODEL.md — happy to reshape it if you'd prefer something different!

potiuk:
Thanks for the careful review, @Claudenw and @ottlinger — really appreciate it. All six threads are addressed in a6978b8 and resolved:

- ✅ `https` for the license URL
- ✅ ASF license headers added to `AGENTS.md` and `SECURITY.md`
- ✅ `SECURITY.md` `TBD` → short security-posture summary + a **Reporting a Vulnerability** section (points at security@apache.org; detailed model stays in `THREAT_MODEL.md`)
- ✅ §3 write-mode note now spells out that `--addLicense` inserts text from the operator-controlled license definition, not the scanned tree (thanks Claude for the "controlled input file" framing)
- ✅ the Maven/Ant/CLI "generated front-ends" point was already captured in the §2 note — pointed there

Should be good to go now — shout if anything else stands out. Thanks again! 🙏

potiuk:
BTW. @Claudenw Usually (in order to save tokens) we either use no licence or we use short SPDX form of it - and usually exclude the AGENTS.md/SECURITY.md from release - to be 100% in line with expectation of full licences in released source packages. 

We had a lengthy discussion about it - without complete conclusion, but the discussion seemed to converge on "If you are not releasing the .md files, it's fine to have no licence or SPDX or some other short version of it. This is generally OK - according to https://www.apache.org/legal/src-headers.html#is-a-short-form-of-the-source-header-available - where one such example short form is allowed in justified cases, while SPDX seems to be another good form. and we usually exclude those files in .rat excludes and ignore on source releases.

Long thread about it (with my assumptions summarized) can be found here - https://lists.apache.org/thread/j1tn63r2lf13v3d1tnnqff8fkcl4nx53

If you wish - we can use either approach.

Greeting from NY Claude :)

THREAT_MODEL.md
potiuk: Good question, Claude — yes, #679 does touch this. The §5a / §8 #2 text already records the hardening (RAT disables external entities, plus your #679 DOCTYPE work), but the data-flow diagram and the input/residual tables were still labelling XXE a bare "surface". I have annotated all three (bb380928) with "mitigated: external entities disabled, RAT-560/#679" so the diagram lines up with §5a/§8. Thanks for flagging the inconsistency!

AGENTS.md
Claudenw, line 1: It looks like AGENTS.md is now just the license.  I don't think that was intended.
If so they should we not just delete the AGENTS.md file?

Claudenw (changes requested):
AGENTS.md looks broken

AGENTS.md
potiuk: Ups.

potiuk: Fixed in `3fee61d` — the earlier "add license header" commit accidentally dropped the whole `AGENTS.md` body, leaving just the comment. Restored the original Agent guidance / CODE STYLE / BEST PRACTICES / COMMUNICATION / RESTRICTIONS sections under the header, and added a SECURITY section wiring `AGENTS.md` → `SECURITY.md` → `THREAT_MODEL.md` (the point of the PR). Sorry Claude :)

SECURITY.md
Claudenw: FYI: RAT picks up the SPDX header so `SPDX-License-Idenfifier: Apache-2.0`  will be picked up expected.

Claudenw (approved):

PR #685, opened 2026-06-28
RAT-562: Use MutableInt in ClaimValidator

Claudenw:
Switch to MutableInt in ClaimValidator to avoid boxing/unboxing

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/config/results/ClaimValidator.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/config/results/ClaimValidator.java
@@ -76,14 +76,18 @@ public final class ClaimValidator {
             DefaultLog.getInstance().warn("`null` passed as argument to setMax() -- ignoring");
             return;
         }
-        MutableInt maxValue = max.computeIfPresent(counter, (k, v) -> {
-            v.setValue(value < 0 ? Integer.MAX_VALUE : value);
-            return v; });
-        min.computeIfPresent(counter, (k, v) ->  {
-            if (v.intValue() > maxValue.intValue()) {
-                v.setValue(maxValue.intValue());
+        MutableInt maxValue = max.compute(counter, (k, v) -> {
+            MutableInt result = v == null ? new MutableInt(k.getDefaultMaxValue()) : v;
+            result.setValue(value < 0 ? Integer.MAX_VALUE : value);
+            return result;
+        });
+        min.compute(counter, (k, v) -> {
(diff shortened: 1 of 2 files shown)


ottlinger:
@Claudenw not sure if my change changed the semantics (computeIfPresent) .... I stumbled into this change when I saw some NPE warnings within ClaimValidator. WDYT? Is the assertion that setting a max value sets a min value implicitly?

Not sure if you can imagine a test that runs into the remaining NPEs or if the issue is theoretical only ;)

Thanks

apache-rat-core/src/main/java/org/apache/rat/config/results/ClaimValidator.java
```diff
@@ -71,10 +71,10 @@ public boolean hasErrors() {
      * @param value the value to set. A negative value specifies no maximum value.
      */
     public void setMax(final ClaimStatistic.Counter counter, final int value) {
-        MutableInt maxValue = max.compute(counter, (k, v) -> {
+        MutableInt maxValue = max.computeIfPresent(counter, (k, v) -> {
```
Claudenw: My reading of computeIfPresent in light of computeIfAbsent and compute is that

- computeIfPresent computes a new value if the key is found in the map.
- computeIfAbsent computes a new value if the key is not found in the map.
- compute just computes the value but the BiFunction has to deal with null values.

I suppose the thing to do here is check for null `counter` argument, log a warning, and exit.

I will add the null check tomorrow.

Claudenw:
@ottlinger please review and merge if you see fit.

apache-rat-core/src/main/java/org/apache/rat/config/results/ClaimValidator.java
```diff
@@ -71,10 +72,14 @@ public boolean hasErrors() {
...
+        if (counter == null) {
+            DefaultLog.getInstance().warn("`null` passed as argument to setMax() -- ignoring");
+            return;
+        }
+        MutableInt maxValue = max.computeIfPresent(counter, (k, v) -> {
             v.setValue(value < 0 ? Integer.MAX_VALUE : value);
             return v; });
-        min.compute(counter, (k, v) ->  {
+        min.computeIfPresent(counter, (k, v) ->  {
             if (v.intValue() > maxValue.intValue()) {
```
ottlinger: @Claudenw the IDE marks this line as possible NPE for maxValue.intValue due to the computerIfPresent before.

ottlinger (approved):
LGTM - thanks for fixing the possibleNPE-related things

PR #686, opened 2026-06-28
RAT-563: Rename IReportable and IReportableListWalker

Claudenw:
Rename IReportable and IReportableListWalker

IReportable is an interface 

IReportableListWalker is a class with a builder.  The name would indicate an interface but that is because it is walking lists of IReportable.

This changes is an internal change that rename IReportable to Reportable and IReportableListWalker with ReportableListWalker


ottlinger (approved):
LGTM - feel free to merge if you are okay with my changes. Thanks for the cleanup.

PR #687, opened 2026-06-28
RAT-564: make LicenseSetFactory return unmodifiable licenses sets

Claudenw:
LicenseSetFactory should return an unmodifiable set of Licenses to prevent accidental modification during processing.

This change does that and signals the unmodifiable state of the result.

This should reduce spotbugs and SonarQube errors.


ottlinger (approved):
LGTM - feel free to merge if you are okay with my changes, @Claudenw

PR #695, opened 2026-07-02
RAT-568: Create IODescriptor

Claudenw:
create and utilize an IODescriptor that will associate a name with the current IOSupplier that is in use.

This change is required by the UI testing changes that are coming shortly.

### DIFF
--- a/apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
+++ b/apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
@@ -20,6 +20,7 @@ package org.apache.rat;
 
 import java.io.File;
 import java.io.FileFilter;
+import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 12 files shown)


apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -843,4 +866,46 @@ public void validate(final Consumer<String> logger) {
...
+        }
+
+        // INPUT CONSTRUCTORS
+
+        /**
+         * Creates an input IODescriptor for the file.  Does not modify for working directory.
+         * @param file the file to open.
+         * @return the Input IODescriptor.
+         */
+        static IODescriptor<InputStream> input(final File file) {
```
ottlinger, line 906: I assume some of these (currently unused methods) will come in handy with new PRs, right?

ottlinger (approved):
Feel free to merge if you are okay with my tiny adjustments. Thanks. @Claudenw

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
Claudenw: Yes, they are in the next PR I belive.

PR #697, opened 2026-07-06
RAT-570: Create Reporter.Output object

Claudenw:
**This change requires RAT-568** (now merged)

This change is to defines a `Reporter.Output` class that contains the configuration that was used for the execution, the XML document that was produced,  and the ClaimStatistic that captures the numbers various types of files, licenses, license categories and other statistics from the run.

The `Reporter.execute()` will create a `Reporter.Output` and will not automatically format the output as specified in the configuration.  Formatting output becomes a second step.

`Reporter.Output` will have 2 methods to format the output 

 * `format(ReportConfiguration)` - will use the output and stylesheet defined in the configuration to format the output.
 * `format(IOSupplier stylesheet, IOSupplier output)` - will apply an arbitrary stylesheet against the contained document and send the output to the location specified.

This change will make it easier for the Maven version to generate the document in the Mojo.check class and the final output in the Mojo.report class without regenerating the report. 

The `Reporter.Output` will be immutable and constructed using a builder pattern.

The builder will be able to read the 3 components (Configuration, XML document,  and ClaimStatistic) from input streams.  Each of those components will be extended to include a Serde implementation to write to, and restore from and IO stream.


apache-rat-core/src/main/java/org/apache/rat/config/exclusion/ExclusionProcessor.java
```diff
@@ -337,4 +395,91 @@ private void extractPaths(final MatcherSet.Builder matcherBuilder) {
         }
     }
 
+    /**
+     * Serializes and deserializes the ExclusionProcessor to XML document
+     */
+    public class Serde {
```
ottlinger: Could we rename this class as Serde does not seem to be a natural word ....
what about ExclusionProcessorXmlConverter or ExclusionProcessorXmlMapper?

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
```diff
@@ -113,6 +123,9 @@ public String displayName() {
     /** Map of counter type to value */
     private final ConcurrentHashMap<ClaimStatistic.Counter, IntCounter> counterMap = new ConcurrentHashMap<>();
 
+    public Serde serde() {
```
ottlinger: I'd prefer to rename this class as well as Serde is rather generic

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -183,6 +196,10 @@ public ReportConfiguration() {
         reportables = new ArrayList<>();
     }
 
+    public Serde serde() {
```
ottlinger: I'd prefer a more specific class name as suggested above.

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
```diff
@@ -242,6 +259,16 @@ public ReportableListWalker.Builder getSources() {
         return builder;
     }
 
+    // for testing access
+    Iterable<File> sources() {
+        return sources;
+    }
+
+    // for testing access
```
ottlinger, line 276: Is there something like Guava's @VisibleForTesting somewhere available in Commons? The more I read these comments, the more I prefer this annotation ;)

apache-rat-core/src/test/java/org/apache/rat/utils/StandardXmlFactoryTests.java
```diff
@@ -18,22 +18,48 @@
...
+import java.nio.charset.StandardCharsets;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
  * Tests for StandardXmlFactory.
  */
-class StandardXmlFactoryTests {
+public class StandardXmlFactoryTests {
```
ottlinger: Is the visibility change important for future PRs?

Claudenw:
Yes serde is serializer-deserializer

LinkedIn: http://www.linkedin.com/in/claudewarren

Claudenw:
@ottlinger My plan is to fix the Maven and Ant issues when they get moved to separate projects.  There will be lots of cleanup there as the code we are running for Maven is a bit old.

ottlinger (changes requested):
LGTM, except for the naming Serde. I'd prefer a named mapper class. WDYT?

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/ExclusionProcessor.java
Claudenw: There are several cases where Serde is the name of the class that serializes and deserializes.

There are some common references:

- https://en.wikipedia.org/wiki/SerDes (though this one has an 's' at the end)
- Rust has a SerDe library.
- Hive has reference to several SerDe implementations: https://cwiki.apache.org/confluence/spaces/Hive/pages/27362059/SerDe
- Kafka streams uses them: https://kafka.apache.org/42/streams/developer-guide/datatypes/
- Spring uses serde: https://docs.spring.io/spring-cloud-stream/reference/kafka/kafka-streams-binder/record-serialization-and-deserialization.html

Several of these call the class/interface `SerDes` some are `Serde`.  I feel that it is a common enough abbreviation that we can use it.  I am open to changing the class name to `SerDe` or `SerDes`

apache-rat-core/src/main/java/org/apache/rat/report/claim/ClaimStatistic.java
Claudenw: The class name is `ClaimStatistic.Serde`.   In light of the comment above is this specific enough?

apache-rat-core/src/main/java/org/apache/rat/ReportConfiguration.java
Claudenw: We could add the guava library to pick up the @VisibleForTesting.  But I think we should do that as a single PR.

apache-rat-core/src/test/java/org/apache/rat/utils/StandardXmlFactoryTests.java
Claudenw: the `simpleDocument` method is used in `OutputTest`.  If hte class is not public the OutputTest can not find the method.

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/ExclusionProcessor.java
ottlinger: I'll rename the classes to SerDes then.

Should we change the builder methods as well or is` SerDes.serde();` ok?

Personally I think `SerDes.serDes();` looks strange to me. 

Or would `build()` be clearer here or `create()`?

WDYT? @Claudenw

Claudenw: Wait.  The outer class in the case above is ExclusionProcessor.  It has a method `public Serde getSerde()` which returns an instance of `ExlcusionProcessor.Serde`.  So we change the name of `ExclusionProcessor.Serde` to `ExclusiongProcessor.SerDes` and the method signature to `public SerDes getSerdes()`?

There should be no `SerDes.serDes()`.  the SerDes should implement a `serializer()` and a `deserializer.()` method.

What am I missing?

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/ExclusionProcessor.java
```diff
@@ -82,8 +82,8 @@ public ExclusionProcessor() {
         excludedCollections = new HashSet<>();
     }
 
-    public Serde serde() {
-        return new Serde();
+    public SerDes serde() {
```
Claudenw: Do you want to change the method from `serde()` to `serDes()`?

apache-rat-core/src/main/java/org/apache/rat/config/exclusion/ExclusionProcessor.java
ottlinger: @Claudenw feel free to merge if you are okay. Personally I think "serDes()" is a bit clunky, but no need to argue any longer ;)

ottlinger (approved):
Thanks

ottlinger:
The new SonarCloud issues are related to using the non-deprecated injection annotations. The one concerning the unused field is not really clear to me as the sink is set to null in one method, while remains in the other. I'm not sure if that changes something within the maven-site-plugin-run, therefore would keep the code as it is.

Claudenw:
@ottlinger I think this is OK to merge now.  If you approve please merge.

PR #698, opened 2026-07-08
RAT-572: Maven build improvements after dropping JDK8 in RAT 0.18

ctubbsii:
RAT-478 updated the Java version to 17 for the minimal build JDK and the target runtime JRE, but left some things in place that are no longer needed. This cleans some of that up.

* Replace `mavenVersion` property with `minimalMavenBuildVersion` property defined in ASF parent POM
* Remove unneeded managed plugin versions and version properties from plugins defined to the same or newer version in the ASF parent POM
* Remove other properties and plugin config that use those properties (like `maven.compiler.*`) when they can be inherited from the parent POM instead
* Remove unused animal-sniffer-maven-plugin (no longer needed with JDK 9 and later)
* Override `minimalJavaBuildVersion` property from ASF parent POM so it can be more easily managed separately from the `javaVersion` (target version) property
* Move Eclipse-specific m2e lifecycle configuration to profile that is activated only in Eclipse, to make it easier to manage separately from actual plugin configuration
* Remove redundant Maven version check from enforcer plugin (already defined in ASF parent POM)
* Fix `<license>` element in the POM to correctly name the Apache license using its SPDX identifier and remove unused generic comment that appears to have been copied from an example configuration
* Remove unneeded maven.compiler.release profile, since it was only needed to support building using JDK8, which is no longer possible

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,27 +45,25 @@ agnostic home for software distribution comprehension and audit tools.
   <properties>
     <sonar.organization>apache</sonar.organization>
     <sonar.host.url>https://sonarcloud.io</sonar.host.url>
-    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- taken from https://maven.apache.org/guides/mini/guide-reproducible-builds.html, updated by maven-release-plugin -->
     <project.build.outputTimestamp>2026-03-11T23:25:45Z</project.build.outputTimestamp>
     <ant.version>1.10.17</ant.version>
     <assertj.version>4.0.0-M1</assertj.version>
+    <!-- this is the target Java version that this project supports at runtime -->
     <javaVersion>17</javaVersion>
     <tika.version>3.3.1</tika.version>
     <mockito.version>5.23.0</mockito.version>
-    <maven.compiler.source>${javaVersion}</maven.compiler.source>
-    <maven.compiler.target>${javaVersion}</maven.compiler.target>
     <!-- This is the version of Maven required to use the RAT Maven Plugin -->
     <mavenMinVersion>3.9</mavenMinVersion>
-    <mavenVersion>3.9.16</mavenVersion>
+    <!-- minimal versions to build this projects, enforced in parent POM by enforcer plugin -->
+    <minimalJavaBuildVersion>17</minimalJavaBuildVersion>
+    <minimalMavenBuildVersion>3.9.16</minimalMavenBuildVersion>
     <creadur.jira.id>RAT</creadur.jira.id>
     <velocity.core.version>2.4.1</velocity.core.version>
     <velocity.tools.version>3.1</velocity.tools.version>
     <!-- maven plugin versions -->
     <mavenPluginTestingVersion>3.5.1</mavenPluginTestingVersion>
-    <mavenPluginPluginVersion>3.15.2</mavenPluginPluginVersion>
     <mavenChangesVersion>3.0.0-M3</mavenChangesVersion>
-    <mavenJavadocPluginVersion>3.12.0</mavenJavadocPluginVersion>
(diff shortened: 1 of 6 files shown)


Claudenw (changes requested):
@ctubbsii 

As I read through these changes it looks like you are removing our specific versions for the versions in ASF parent.  

Most of these seem fine to me, however, I am a bit concerned about the maven build changes.  We build a maven plugin so I think we need the maven plugin plugin bits.

Have you tries to build the entire package from scratch using these mods?
Do you have experience building a maven plugin?

I like where this change is heading, I am just a bit nervous about some of the more obscure bits.

Not really requesting changes but want to hold merging until I am comfortable.

Claude

pom.xml
```diff
@@ -45,27 +45,25 @@ agnostic home for software distribution comprehension and audit tools.
...
+    <minimalJavaBuildVersion>17</minimalJavaBuildVersion>
+    <minimalMavenBuildVersion>3.9.16</minimalMavenBuildVersion>
     <creadur.jira.id>RAT</creadur.jira.id>
     <velocity.core.version>2.4.1</velocity.core.version>
     <velocity.tools.version>3.1</velocity.tools.version>
     <!-- maven plugin versions -->
     <mavenPluginTestingVersion>3.5.1</mavenPluginTestingVersion>
-    <mavenPluginPluginVersion>3.15.2</mavenPluginPluginVersion>
     <mavenChangesVersion>3.0.0-M3</mavenChangesVersion>
-    <mavenJavadocPluginVersion>3.12.0</mavenJavadocPluginVersion>
```
ottlinger, line 68: Due to RAT-497 we defined Javadoc plugin version in our own tree ..... but we can rely on the ASF parent for that as these older JDK versions do not receive any more support anymore.

ctubbsii:
@Claudenw wrote:
> Most of these seem fine to me, however, I am a bit concerned about the maven build changes. We build a maven plugin so I think we need the maven plugin plugin bits.

I didn't remove anything that wasn't already defined elsewhere. Those were defined in the `pluginManagement` section solely for version management. They were redundant with what was declared in the ASF parent POM that this project uses. So, they were serving no purpose. No plugin execution or usage was removed, only the redundant version management.

> Have you tries to build the entire package from scratch using these mods? Do you have experience building a maven plugin?

I built the entire package using `mvn verify` and all the GitHub Actions CI checks also performed the build. This project has several integration tests that make use of the plugin in a small test project to validate that it still works. Aside from that, I don't know what you mean about "build the entire package from scratch".

Also yes, I have experience building maven plugins. I have created several, and help maintain some that are widely used (like [formatter-maven-plugin](https://code.revelc.net/formatter-maven-plugin/plugin-info.html)).

> I like where this change is heading, I am just a bit nervous about some of the more obscure bits.

Which bits? I'm happy to explain any bits you don't understand.

> Not really requesting changes but want to hold merging until I am comfortable.

Please let me know what it would take for you to get comfortable with the changes.

Claudenw:
I am glad to see you ran the build.  I will dive deeper this weekend.  But from 10K feet it looks good.

pom.xml
ctubbsii: These properties were previously necessary to use newer versions of these plugins to override the versions defined in the ASF parent POM. Since then, the parent POM was updated, and uses these newer versions already, so these are no longer needed. New patches can override the versions again if it ever becomes necessary to override the version from the parent POM again, but right now, that is not necessary. However, if it does become necessary, the parent POM makes it easier to override the version by managing the plugin versions with properties, so you don't need to make your own properties and pluginManagment entries to merely override a version.

ottlinger (approved):
LGTM

ottlinger:
@Claudenw feel free to merge; @ctubbsii thanks for taking a look into the situation.

Claudenw (approved):

PR #703, opened 2026-07-26
Convert file arg to documentname

Claudenw:
Make file argument type return documentname so that UI processing can properly reference the relative directory.

part of RAT-541


Claudenw:
Error in creation

PR #713, opened 2026-08-18
[RAT-574] recursive regex walking fix

rmannibucau:


Claudenw (approved):
Thank you for your submission.

I like the approach.  However, I have been thinking that we need a better way to do regex like searches in large text.

Until we can come up with such a model this will do nicely.

Thank you again.

PR #730, opened 2026-09-17
Fix sonar cloud warnings and get a green quality build on RAT

ottlinger:
* [x] Remove debugging var usage in test
* [x] Use java.time instead of java.date
* [x] NOSONAR for Injection in MOJO
* [ ] [ThreadLocal cleanup](https://sonarcloud.io/project/issues?id=apache_creadur-rat&open=AZ-vnccXyTDlcsX8JlQc&s=IMPACT_RANK&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=BUG) ?! remove() cannot be called within the class itself

PR #731, opened 2026-09-19
Fix remaining sonarcloud warnings

ottlinger:
Followup of #730:
* replace assertion in test as suggested by sonar analysis
* NOSONAR ThreadLocal warnings to call remove as classes are handed to other classes in RAT


Claudenw (approved):
LGTM.

